PR #11, opened 2021-07-09
task 73

amirrezamirzaei:
adding my first task.
[paper](https://arxiv.org/abs/1811.00937)
dataset: [CommonsenseQA](https://www.tau-nlp.org/commonsenseqa)

### DIFF
--- a/tasks/task073_CommonsenseQA_answer_generation.json
+++ b/tasks/task073_CommonsenseQA_answer_generation.json
@@ -26,10 +26,7 @@
         },
         {
             "input": "What island country is ferret popular?\n(A)own home (B)north carolina (C)great britain (D)hutch (E)outdoors",
-            "output": [
-                "B",
-                "C"
-            ],
+            "output": "B,C",
             "explanation": "Only one of the following options ('A', 'B', 'C', 'D' or 'E') is allowed to be the answer. Multiple options can not be correct. The correct answer here would have been 'C'."
         }
     ],


swarooprm (changes requested):
@amirrezamirzaei, thanks for creating the pull request. Good start! Some feedbacks:
1. "Categories" field should have value "Answer Generation" and not just "Generation"
2. Create negative examples (one way to create negative example is to generate an output which is not A, B, C or D) e.g. https://instructions.apps.allenai.org/dataset_viewer?file=subtask043_essential_terms_answering_incomplete_questions.json
3.  "Definition" field looks skinny. Look at task001, "Definition" should be dense and contain "Things to Avoid" and "Emphasis and Caution" similar to task001, task003 etc. We usually get all these information from the crowdsourcing template of the corresponding dataset. So, may be you can contact authors of commonsenseQA and get access to the crowdsourcing template, as suggested in step 1 of our guideline. 

Feel free to email us if you face any difficulty.

swarooprm (commented):
Good job @amirrezamirzaei on your first task!

My review: Accept!
Waiting for @danyaljj's review

You can also create more tasks from the same commonsenseQA dataset: e.g. incorrect answer generation (e.g. https://instructions.apps.allenai.org/dataset_viewer?file=subtask005_mctaco_wrong_answer_generation_event_duration.json)

danyaljj:
Looks good to me! Merging!

PR #15, opened 2021-07-12
Added 4 tasks using SPLASH dataset and 1 using the CoNaLa dataset.

kurbster:


tasks/task075_splash_question_to_sql.json
```diff
@@ -0,0 +1,12243 @@
+{
+    "Contributors": [
+        "Kirby Kuznia"
+    ],
+    "Categories": [
+        "Answer Generation"
+    ],
+    "Definition": "In this task you are expected to write an SQL query that will return the data asked for in the question.",
```
danyaljj: The definition should enough for an average person to do the task. 
So, I think we need to introduce & define the SQL operators that will be used for this answering this task (SELECT, FROM, AS, WHERE, *, LIMIT). Basically, a short introduction to SQL syntax.

danyaljj (changes requested):
Thanks, @kurbster! 

The key change that we need here is to make sure that all the tasks come with self-contained definitions. Basically, one who is not closely familiar with SQL should be able to read the "definition" and perform the task.

swarooprm (changes requested):
Looks good. One comment: The Definition field needs to be more dense and contain all the information required in understanding SQL.

tasks/task075_splash_question_to_sql.json
kurbster: I mention basic syntax for the following, `SELECT, FROM, AS, JOIN, ON, WHERE, GROUP BY, HAVING, ORDER BY, DESC, ASC, LIMIT`. I just appended the explanation to the end of every SQL task. Let me know if it's too verbose or too brief.

danyaljj:
Thanks! Looks much better! 

Few small issues: 

- The last part of this sentence (`\"ON\" condition statement`) is not clear to me: `... If you want to select data from multiple tables you need to use the \"JOIN\" table \"ON\" condition statement` 

- I think we need to define what `*` too, since I see it in many examples. 

- In "task076_splash_correcting_sql_mistake": 
      -  "it should min salary" -> missing "be"? 
      - "Swap avg average with average capacity ." -> should we use quotations to make it a bit less ambiguous? Say: `Swap 'avg (Average)' with 'avg (Capacity)' .` Alternatively, it could be more English-like if it was phrased: `Swap the average of  "average" with an average of "capacity" .`

- I think the task types should be more accurate (something along the line of `symbolic-query-generation`). I'll let @swarooprm comment on that. 

- Could you resolve the conflict with the master branch?

danyaljj:
Thanks! Merging!

danyaljj:

PR #16, opened 2021-07-12
2 subtasks from PIQA dataset

kuntalkumarpal:
1. Correct Answer Generation Given a goal
2. Incorrect Answer Generation, Given a goal

Randomly considered 6.5K samples.


kuntalkumarpal:
Need some changes

PR #22, opened 2021-07-17
babi dataset task 1- 3 NI tasks (question generation, answer generation, identify supporting fact for QA)

shailaja183:


tasks/task082_babi_t1_single_supporting_fact_question_generation.json
```diff
@@ -0,0 +1,6053 @@
+{
+    "Definition": "In this task, you will be given a passage consisting of set of facts. Among these facts, choose any one as a supporting fact, and rest of them will be irrelevant facts. The task is to create a question of form 'Where is <person_name>?' that is answerable from the supporting fact. Avoid creating questions that are unanswerable from all of the facts.",
```
danyaljj: I think the sentence `Among these facts, choose any one as a supporting fact, and the rest of them will be irrelevant facts.` is not adding much to this task. If that's the case, let's drop it.

tasks/task082_babi_t1_single_supporting_fact_question_generation.json
```diff
@@ -0,0 +1,6053 @@
...
+            "output": [
+                "Where is Daniel?"
+            ]
+        },
+        {
+            "input": "Passage: Mary moved to the bathroom. John went to the hallway. Daniel went back to the hallway. Sandra moved to the garden. John moved to the office. Sandra journeyed to the bathroom. Mary moved to the hallway. Daniel travelled to the office.",
+            "output": [
+                "Where is Daniel?"
+            ]
+        },
```
danyaljj, line 79: Is it possible to augment each instance with all the possible outputs? 

For example, in this case, the output should be: 
```
            "output": [
                "Where is Daniel?", "Where is Mary?", "Where is John?", "Where is Daniel?", "Where is Sandra?", ...
            ]```

Comment (danyaljj) [reply tasks/task082_babi_t1_single_supporting_fact_question_generation.json:79]: This shouldn't be difficult since the dataset is using a closed set of names, I think.

File: tasks/task083_babi_t1_single_supporting_fact_answer_generation.json [hunk 3]
```diff
@@ -0,0 +1,6058 @@
+{
```
danyaljj, line 1: This task looks good to me.

tasks/task084_babi_t1_single_supporting_fact_identify_relevant_fact.json
```diff
@@ -0,0 +1,6048 @@
+{
```
danyaljj: This task looks good to me.

tasks/task082_babi_t1_single_supporting_fact_question_generation.json
shailaja183: Updated the instances with all possible questions for a given passage in the new commit. However, I have listed only unique questions (ex. if John appears multiple times in the passage, there will be only one question- Where is John?). I am not sure what would be the benefit of repeating questions shown in your example above.

tasks/task082_babi_t1_single_supporting_fact_question_generation.json
shailaja183: Updated the definition in the new commit, let me know if that looks fine.

tasks/task082_babi_t1_single_supporting_fact_question_generation.json
danyaljj: Unique question is the way to go. The list really shows all the unique outputs.

danyaljj:
Looks good to me! Assuming that @swarooprm has reviewed it earlier, merging!

danyaljj:

PR #29, opened 2021-07-18
Adding tasks using examples from CoNaLa dataset.

kurbster:
Task 096 I created on my own, whereas every other dataset was created by finding an example in the CoNaLa dataset then generating data through python.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -73,9 +73,9 @@ Name | Summary | Category
 `task077_splash_explanation_to_sql` | Generate an SQL statement based on a description of what the SQL statement does. | Structured Query Generation
 `task078_splash_sql_to_explanation` | Give a natural language description of what a given SQL statement is doing. | Structured Query Classification
 `task079_conala_concat_strings` | Given a list of strings concatenate them to form one string | Answer Generation
-`task093_conala_normalize_list` | Given a list of numbers normalize the list such that the result adds to 1 | Answer Generation, Mathematics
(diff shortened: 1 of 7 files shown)


danyaljj:
Thanks! 

Several small suggestions: 

 - Simplify the phrasing a bit: 
```
In this task you will be given a list of numbers and you need to normalize them. Normalization in this case means that the output list should sum to the number one and every value in the input list should have the same weight in the output list.
```
How about this?: 
```
In this task, you will be given a list of numbers. The goal is to multiply all the numbers in the list by a constant such that the sum of the resulting numbers is one. 
```
 - `This is a bad example.` --> `This is an incorrect output`
 - Note that not all your numbers have 3 
 - in all the tasks, for the negative examples, add a short sentence/phrase that indicates the fix (i.e., how to turn them into positive examples).

swarooprm (commented):
Looking good! I am happy that size of the list is a variable in instances, this is a good way to evaluate "scale invariance" in models.

Some comments:
1. Categories: Instead of "Mathematics", you can add "arithmetic". task097 category needs to be updated.
2. I would add formula of each task in the definition e.g. in case of average, you can say average can be found by summing numbers in the list and then dividing it by the number of elements in that list.

danyaljj (approved):
Looks good to me! 

@swarooprm  feel free to merge when happy.

swarooprm (approved):

danyaljj:

kurbster:
@danyaljj I can look at the feedback today. Should I address all the comments then create another PR?

PR #14, opened 2021-07-10
Shailaja tasks

shailaja183:
Pull request for including task074_squad1.1_question_generation

PR #32, opened 2021-07-19
squad2.0

shailaja183:
3 NI tasks (question generation, answerable/unanswerable question classification, answer generation)

### DIFF
--- a/tasks/task082_babi_t1_single_supporting_fact_question_generation.json
+++ b/tasks/task082_babi_t1_single_supporting_fact_question_generation.json
@@ -46,44 +46,44 @@
         {
             "input": "Passage: Mary moved to the bathroom. John went to the hallway.",
             "output": [
-                "Where is John?",
-                "Where is Mary?"
+                "Where is Mary?",
+                "Where is John?"
             ]
         },
         {
             "input": "Passage: Mary moved to the bathroom. John went to the hallway. Daniel went back to the hallway. Sandra moved to the garden.",
             "output": [
-                "Where is John?",
-                "Where is Daniel?",
                 "Where is Mary?",
-                "Where is Sandra?"
+                "Where is John?",
+                "Where is Sandra?",
+                "Where is Daniel?"
             ]
         },
         {
             "input": "Passage: Mary moved to the bathroom. John went to the hallway. Daniel went back to the hallway. Sandra moved to the garden. John moved to the office. Sandra journeyed to the bathroom.",
             "output": [
-                "Where is John?",
-                "Where is Daniel?",
                 "Where is Mary?",
-                "Where is Sandra?"
+                "Where is John?",
+                "Where is Sandra?",
+                "Where is Daniel?"
             ]
         },
         {
             "input": "Passage: Mary moved to the bathroom. John went to the hallway. Daniel went back to the hallway. Sandra moved to the garden. John moved to the office. Sandra journeyed to the bathroom. Mary moved to the hallway. Daniel travelled to the office.",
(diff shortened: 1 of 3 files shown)


danyaljj:
How are the tasks in #14 different from those here? (except `tasks/task100_squad2.0_determine_answerable_question.json`)

shailaja183:
In squad1.1 (#14), only questions which are answerable from the passage were allowed. 
In suqad2.0, they relaxed this assumption by having unanswerable questions so that a model can learn when an answer is possible and when not.   

In addition to typical question generation instructions, following conditions are added for unanswerability; 
(i) if negation words are added or removed to/from the sentence in the passage 
(ii) if antonyms are used for the words in the passage 
(iii) entity, number, or date is replaced with other entity, number, or date 
(iv) if word or phrase in question is mutually exclusive with something for which an answer is present 
(v) if question asks for the condition that is not satisfied by anything in the passage 
(vi) answer needs information that is not provided in the passage.
 
For answer generation task, model must return empty string if question is unanswerable. For answerable questions, instructions are same as squad1.0.

danyaljj:
- Regard QG: 
According to [the paper](https://arxiv.org/pdf/1806.03822.pdf), they annotated "unanswerable" questions only and mixed them with SQuAD 1.0 questions. I think it makes sense to follow the same setup here (just so that we're faithful to the original construction). 
Screen Shot 2021-07-20 at 4 58 56 PM

shailaja183:
Just to confirm, I should be doing following changes-
1. Question Generation task - Filter out unanswerable questions from squad2.0 and rewrite task description (as per your suggestion above)
2. Unanswerable/Answerable Question Classification - No changes required
3. Answer Generation task - there are two possibilities;
(i) If task is w.r.t. unanswerable questions only, it does not make sense (as output always will be empty string) so simply remove this task 
(ii) If the task involves answerable+unanswerable questions (is already done in the current file I have submitted). 
Let me know which one do you prefer.

danyaljj:
> Question Generation task - Filter out unanswerable questions from squad2.0 and rewrite task description (as per your suggestion above) 

Conversely, Filter out **answerable** questions from squad2.0  (retain "unanswerable" questions). So, the task will be "Unanswerable Question Generation Task". 


> Answer Generation task - there are two possibilities; 
> (i) If task is w.r.t. unanswerable questions only, it does not make sense (as output always will be empty string) so simply remove this task
> (ii) If the task involves answerable+unanswerable questions (is already done in the current file I have submitted).
> Let me know which one do you prefer.

(ii) makes sense.

shailaja183:
Yes, that's what I meant for question generation. Sorry for confusion. I will push the changes.

shailaja183:
I pushed the changes as per our discussion yesterday. Let me know if it looks good. 
Please refer to commit -- fixed question generation task (f9164af) --  for changes

danyaljj:
Thanks! I am happy with the QG task now! 

Few minor suggestions regarding the remainder of the tasks: 

* task100_squad2.0_answerable_unanswerable_question_classification

 - "Carefully read the given passage and a question. "  --> seems redundant, especially the "carefully read" part 
 - In the file, I see 4k "false" instances and 2k "true" instances. Could you make this more balanced? This suggestion applies to the QA task too (have a balanced portion of answerable and unanswerable questions). 


* task101_squad2.0_answer_generation 
- "output is not generated" --> not clear. Do you mean, the expected output must be empty? 
- I am not sure if the content after "Although question might seem relevant to the passage ..." is really adding anything. What would we lose if we drop these?

PR #34, opened 2021-07-22
Task103 facts to story

Mihir3009:
Added new task 103: Facts to Story. Please review.


danyaljj:
Thanks for the PR! 

Few very minor suggestions: 
 - For "Sentence Generation": mention that the concept names are separated by "#". 
 - Merge "There are 3 to 5 concepts in each concept set. " with the first sentence: "you are given concept set (with 3 to 5 concepts". 
- We need to use **all** the provided concepts in the output generation, right? If so, we can create a negative example that uses k-1 concepts (from a concept set of size k).
- I see that some instances have more than one gold output (which is good!) If the dataset has more instances than 3.5k, I would suggest selecting a subset of instances that a higher number of gold outputs per instance. 

Long text generation: 
- Merge "In this task, five ordered key facts are given. All the given facts are expressed in natural language." --> more concise 
- There are a bunch of `<sep><cls>` tokens; drop. 

Please update the readme of task names as well.

Mihir3009:
I updated the task 102 file - Sentence Generation. Please review.

I have a question regarding Task 103 - Long text generation:
When you say "There are a bunch of "" tokens; drop.". Can you please explain it more? I didn’t get the point. So that I can review and resubmit.

danyaljj:
> When you say "There are a bunch of "" tokens; drop.". Can you please explain it more? I didn’t get the point. So that I can review and resubmit. 

Sorry, I miss-typed. There are a bunch of non-natural(`<cls>` or other special tokens) that we should drop.

Mihir3009:
Got it.

I updated both the files according to your suggestions. Please review. Thanks!

swarooprm (approved):
Looks good to me! Just a minor suggestion:
Change the file names that indicate the source dataset and the category e.g. if source for both tasks are 'commongen', then task name should be taskNNN<source>_<category>.json (e.g. 'task001_quoref_question_generation.json')

Mihir3009:
I renamed the file and uploaded the final ones.

swarooprm:
Accept from my side!

PR #51, opened 2021-07-26
ASSET two tasks 111 and 112

Palipoor:
I create two tasks with the dataset ASSET. This is my first time doing this and I would appereciate your feedback. Thanks.

### DIFF
--- a/tasks/task078_splash_sql_to_explanation.json
+++ b/tasks/task078_splash_sql_to_explanation.json
@@ -92,12 +92,6 @@
             ]
         },
         {
-            "input": "SELECT city FROM Addresses GROUP BY city ORDER BY Count ( * ) Desc LIMIT 1",
-            "output": [
-                "Step 1: find the number of rows of each value of city in Addresses table.\nStep 2: find city of Addresses table with largest value in the results of step 1"
-            ]
-        },
-        {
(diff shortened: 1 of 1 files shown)


Palipoor:
Oh I see. I will fix the problem.

danyaljj:
Thanks! 

- Do you think `for non-native English speakers` is necessary? Basically, I am wondering whether native speakers of English can objectively solve these tasks, in which case, the mention of `non-native speakers` is redundant. 
- `have unnecessary   phrases`: redundant space before `phrases`  
- In the task definition, define what is the expected output. `You are expected to output "Text one"  if the first sentence is simpler. Otherwise, ....` 
- After PR #53 is merged, please pull and merge with the main branch. It should trigger a bunch of new tests.

Palipoor:
@danyaljj Thanks for the feedback. Sorry if I had many mistakes, it was my first time doning this.
The reason I put "non-native speakers" is the task description in ASSET dataset's mturk templates had it. I will remove it.
I will also fix other problems. thanks.

danyaljj:
> The reason I put "non-native speakers" is the task description in ASSET dataset's mturk templates had it.  

If that's how the original task was phrased, let's keep it. We want to remain as faithful as possible to the original instructions.

Palipoor:
At first I fixed some duplicate issues with other tasks but then I realized I'll encounter conflicts with the master branch, so I just went back.

danyaljj (approved):
Thanks! It's green from my side! ✅ 
@swarooprm feel free to merge when you're happy with the task.

swarooprm (approved):
Looks good! Merging.

Issue #55, opened 2021-07-27
Randomization issue in tests


swarooprm:
@nrjvarshney reports that there is possibly some randomization in the testing script. Sometimes an error appears, sometimes it does not. Also, sometimes all test passes in local, but on the repo, the test fails.

danyaljj:
it's due to this line: 
https://github.com/allenai/natural-instructions-expansion/blob/master/src/test_all.py#L58-L59 
You can increase `500` to a larger number to cover more cases.

PR #41, opened 2021-07-23
task108 abuse detection

Maitreyapatel:
Added new task 111 for abuse detection. Please review it.

PR #56, opened 2021-07-27
removing randomization (issue #55)

swarooprm:


tasks/task075_squad1.1_answer_generation.json
```diff
@@ -11476,12 +11476,6 @@
                 "geography, climate, and flora"
             ]
         },
-        {
-            "input": "Passage: In their attempt to ensure white supremacy decades after emancipation, in the early 20th century, most southern states created laws based on the one-drop rule, defining as black, persons with any known African ancestry. This was a stricter interpretation than what had prevailed in the 19th century; it ignored the many mixed families in the state and went against commonly accepted social rules of judging a person by appearance and association. Some courts called it \"the traceable amount rule.\" Anthropologists called it an example of a hypodescent rule, meaning that racially mixed persons were assigned the status of the socially subordinate group. Question: What is the one-drop rule?",
-            "output": [
-                "black, persons with any known African ancestry."
```
danyaljj, line 11482: If this is repeated input, it should be merged with the existing labels, rather than being removed.

swarooprm: done

swarooprm:
Currently, tests run for all tasks, I am fixing in another PR to run test for just the task which is modified. Once this PR gets merged, https://github.com/allenai/natural-instructions-expansion/pull/41 will pass.

src/test_all.py
```diff
@@ -18,63 +18,62 @@
 
 with open("tasks/README.md", 'r') as readmef:
     task_readme_content = " ".join(readmef.readlines())
-
 files = [f for f in listdir(tasks_path) if isfile(join(tasks_path, f))]
+files.sort()
+lastindex=int(files[-1].split('_')[0].split('task')[1])
 for file in files:
```
danyaljj: Since the files are sorted according to the index id, I think it's enough to do `for file in files[70:]` and drop all the `lastindex`-relatd code.

swarooprm: I guess you meant files[lastindex-70:] as files[70:] can be huge once the task count reaches 500 or 1000. The code files[lastindex-70:] ensures that only 70 tasks are getting evaluated and this way testing will take a couple of minutes only.

danyaljj:
I meant files[-70:].


Daniel

danyaljj:
Closes issue #55

src/test_all.py
```diff
@@ -53,28 +52,26 @@
...
+                        raise Exception(f" * Looks like we have a repeated example here! Merge outputs before removing duplicates. :-/ \n {x}\n {y}")
+
             # make sure there are no examples repeated across instances and positive examples
             for x_idx, x in instances:
                 for y_idx in range(len(data['Positive Examples'])):
                     y = data['Positive Examples'][y_idx]
                     if x['input'] == y['input']:
-                        raise Exception(f" * Looks like we have a same example across positive examples and instances! :-/ \n {x}\n {y}")
-            
+                        raise Exception(f" * Looks like we have a same example across positive examples and instances! Merge outputs before removing duplicates. :-/ \n {x}\n {y}")
```
danyaljj: Instead of "Merge outputs ..." it should be "Drop the example from the instances".

danyaljj:
One suggestion, and otherwise, looks good! Merge when happy!

danyaljj (approved):

PR #60, opened 2021-07-29
Task 116 Com2sense task - Commonsense reasoning

Palipoor:
A commonsense reasoning task based on the Com2Sense dataset.
https://arxiv.org/abs/2106.00969


swarooprm:
Please rename your task to task116 since task115 is already submitted in the latest Pull request.

tasks/README.md
```diff
@@ -96,4 +96,5 @@ Name | Summary | Category
 `task102_commongen_sentence_generation` | Given a collection of concepts, use them in a coherent sentence. | Sentence Generation 
 `task103_facts2story_long_text_generation` | Given 5 facts, write a story that incorporates them. | Long Text Generation 
 `task107_splash_question_to_sql` | Generate an SQL statement from a question asking for certain data. | Structured Query Generation
-`task108_ContextualAbuseDetection_classification` | Given a text detect whether it's abusive or not. | Classification
\ No newline at end of file
+`task108_ContextualAbuseDetection_classification` | Given a text detect whether it's abusive or not. | Classification
+`task_116_com2sense_commonsense_reasoning` | Decide whether a sentence is true or not, based on commonsense | Commonsense Reasoning
```
danyaljj: `task_116` -> `task116` 
I would suggest changing "sentence is true" to whether "the provided event is plausible".

Palipoor: That makes sense, thanks.

tasks/task_116_com2sense_commonsense_reasoning.json
```diff
@@ -0,0 +1,12051 @@
+{
+    "Definition": "You will be given one sentence about an everyday event. You need to decide whether the sentence is true or false, based on your commonsense. For true sentences output 'True' and for false sentences output 'False'.",
```
danyaljj: I am not sure what it means for a sentence to be "true" or "false". 
I would suggest changing the phrasing of the instruction, to something along the line of, "you need to decide whether provided story in the following sentence is a plausible scenario and it matches your commonsense".

danyaljj: Similarly, instead of saying "For true sentences output 'True'" (which sounds like a recursive statement), say, "if the given event is a plausible one, output 'True'" and so on.

tasks/task_116_com2sense_commonsense_reasoning.json
```diff
@@ -0,0 +1,12051 @@
...
+            "explanation": "Bath tubs are filled with water and a paper tape will dissolve in a bath tub, so a smart person would not give such an advice."
+        },
+        {
+            "input": "After a year of consistent practice, Jennifer was able to run a marathon.",
+            "output": "Correct",
+            "explanation": "The output should be either 'True' or 'False'."
+        }
+    ],
+    "Source": [
+        "Com2Sense"
```
danyaljj: Expand this a bit, something like this: The instructions and the instance are extracted from the Com2Sense (<URL to the dataset>)

tasks/README.md
danyaljj: For the last column, maybe also add `Classification`, too.

tasks/task_116_com2sense_commonsense_reasoning.json
```diff
@@ -0,0 +1,12051 @@
...
+    ],
+    "Contributors": [
+        "Pegah Alipoormolabashi"
+    ],
+    "Categories": [
+        "Commonsense Reasoning"
```
danyaljj: Add `Classification`, too.

danyaljj:
That's all the thoughts I had. 
After addressing them, @swarooprm feel free to merge when you're happy.

danyaljj:
@Palipoor pinging the thread to make sure you have seen the comments above.

Palipoor:
@danyaljj Yes I have seen them, I was just too busy these days. Will handle them soon.

swarooprm:
Looks good! Merging this task. 
However, I think you can create more tasks from the same dataset. Section 2.2. contain multiple steps in the crowdsourcing process, you can convert some of those to tasks. To clarify, any input-output can constitute a task. A task does not necessarily have to be an end task in a dataset. Feel free to create new PR if you find any new task using this dataset.

PR #66, opened 2021-08-02
Add `task117_spl_translation_en_de` task

Mehrad0711:


swarooprm (commented):
Thanks for creating the PR! This looks good to me. Accept!
Note that you can create more tasks using the same source (https://github.com//stanford-oval//SPL)

danyaljj:
Looks good, thanks!

Mehrad0711:
Hi!
Sure. I can add more translation tasks following the same naming format for other target languages: `taskXXX_spl_translation_en_<tgt_lang>`
Do you prefer one or separate PRs for each task?

swarooprm:
It's better if all tasks for a source are in a single PR.
But it's also ok if they are created via multiple PRs.

Here, since we already merged your first task, please create new PR.

PR #72, opened 2021-08-04
task115_sec_filings_embeddings.json

shau0000:


danyaljj:
The added task file should be inside `tasks` directory: 
https://github.com/allenai/natural-instructions-expansion/tree/master/tasks

swarooprm:
Thanks for the PR! Also the latest task name in the repo is task117, so please name your task as task118.

danyaljj:
It seems to me that the task is well-defined for humans (i.e., humans would not be able to read the task description and solve it to a good extent). If that's the case, we can't accept your submission, unfortunately.

PR #76, opened 2021-08-06
Task 118-119: semeval math challenges 

amirrezamirzaei:
splitting task 104 ([semeval-2019 task10)](https://github.com/allenai/semeval-2019-task-10) into closed and open vocabulary.

### DIFF
--- a/tasks/task104_semeval_2019_task10_closed_vocabulary_mathematical_answer_generation.json
+++ b/tasks/task104_semeval_2019_task10_closed_vocabulary_mathematical_answer_generation.json
@@ -41,18 +41,6 @@
             ]
         },
         {
-            "input": "If k is an integer and \\(k = \\frac { m } { 3 } \\) which of the following must be true? I. m is an even number. II. m is a multiple of 3. III. m is an odd number.\n(A)I only (B)II only (C)III only (D)I and II only (E)I, II, and III",
-            "output": [
-                "B"
-            ]
-        },
-        {
-            "input": "If x = 1 = kx, for which of the following values of k will the equation have NO solution?\n(A)-1 (B)0 (C)1 (D)2 (E)3",
-            "output": [
-                "C"
-            ]
-        },
-        {
             "input": "The sum of two consecutive even integers is \\(t\\). In term of \\(t\\), what is the value of the greater of the two numbers?\n(A)\\(t + 2\\) (B)\\(t + 1\\) (C)\\(\\frac { t } { 2 } \\) (D)\\(\\frac { t + 1 } { 2 } \\) (E)\\(\\frac { t + 2 } { 2 } \\)",
             "output": [
                 "E"
--- a/tasks/task118_semeval_2019_task10_open_vocabulary_mathematical_answer_generation.json
+++ b/tasks/task118_semeval_2019_task10_open_vocabulary_mathematical_answer_generation.json
@@ -23,11 +23,6 @@
     ],
     "Negative Examples": [
         {
-            "input": "John has y dollars to spend on some new CDs from Music Plus, an online record store. He can buy any CDs at the members' price of x dollars each. To be a member, John has to pay a one-time fee of $19. Which of the following expressions represents the number of CDs John can purchase from Music Plus?\n(A)\\(\\frac{xy}{19}\\) (B)\\(\\frac{y + 19}{x}\\) (C)\\(\\frac{2y - x}{19}\\) (D)\\(\\frac{y - 19}{x}\\) (E)\\(\\frac{19 - x}{y}\\)",
-            "output": "I dont know.",
-            "explanation": "Do not generate anything else apart from one of the following characters: 'A', 'B, 'C', 'D', 'E'."
-        },
-        {
             "input": "There are four colors of lottery tickets in a hat. If \\(\\frac { 1 } { 10 } \\) of the tickets in the hat are green, \\(\\frac { 1 } { 2 } \\) are white, \\(\\frac { 1 } { 4 } \\) are blue, and the remaining 30 tickets are pink, what is the number of blue tickets in the hat?\n(A)25 (B)50 (C)75 (D)120 (E)200",
             "output": "B,D",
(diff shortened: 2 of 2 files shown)


swarooprm (commented):
Thanks! Definition of both the tasks look very similar. May be you can add information that are exclusive to the task e.g. in case of task104_semeval_2019_task10_mathematical_answer_generation, you can say something like this "Questions in this task usually contain some variables (e.g. X, Y), equations (e.g. X+Y=3) and some relationships (X is greater than Y). Often you need to find the value of the variable using equations and relationships" and similarly in case of task 118, you can add something that is exclusive to that task.

amirrezamirzaei:
also adding task 119 to this pr because it was made with semeveal-2019 task10 geometric questions. not really sure about because it doesn't contain that many instances (it has 131 instances).

tasks/task118_semeval_2019_task10_open_vocabulary_mathematical_answer_generation.json
```diff
@@ -0,0 +1,1508 @@
...
+    "Contributors": [
+        "Amirreza Mirzaei"
+    ],
+    "Source": [
+        "semeval 2019 task10"
+    ],
+    "Categories": [
+        "Math Question Answering"
+    ],
+    "Definition": "You are given a mathematical question described with an open-ended vocabulary. Questions in this task don't contain mathematical equation but instead involve real-world situation, describing a mathematical problem. you are also given 4 or 5 answer options (associated with \"A\", \"B\", \"C\", \"D\", \"E\"). Things to avoid: 1. Do not generate anything else apart from one of the following characters: 'A', 'B, 'C', 'D', 'E'. Emphasis & Caution: 1. LaTeX mathematical format is used to express equations. 2. Each question is solvable with high school math knowledge. 3. Only give one answer for each question.",
```
danyaljj: I would simplify it:  "Questions in this task involve ..."

danyaljj: > LaTeX mathematical format is used to express equations. 

An average human does not know what "LaTeX mathematical format" is. Reformulate/rephrase this.

danyaljj: `Only give one answer for each question.`  -> 'Give only one answer ...`

tasks/task104_semeval_2019_task10_closed_vocabulary_mathematical_answer_generation.json
```diff
@@ -0,0 +1,4148 @@
...
+    "Contributors": [
+        "Amirreza Mirzaei"
+    ],
+    "Source": [
+        "semeval 2019 task10"
+    ],
+    "Categories": [
+        "Math Question Answering"
+    ],
+    "Definition": "You are given a mathematical question described with a circumscribed mathematical vocabulary. Questions in this task usually contain some variables (e.g. X, Y), equations (e.g. X+Y=3) and some relationships (X is greater than Y). Often you need to find the value of the variable using equations and relationships. you are also given 4 or 5 answer options (associated with \"A\", \"B\", \"C\", \"D\", \"E\"). Things to avoid: 1. Do not generate anything else apart from one of the following characters: 'A', 'B, 'C', 'D', 'E'. Emphasis & Caution: 1. LaTeX mathematical format is used to express equations. 2. Each question is solvable with high school math knowledge. 3. Only give one answer for each question.",
```
danyaljj: > you are also 

capitalization 

>  LaTeX mathematical format is used to express equations. 

Same comment as the next task (unclear what "LaTeX mathematical format " means)..

tasks/task119_semeval_2019_task10_geometric_mathematical_answer_generation.json
```diff
@@ -0,0 +1,824 @@
...
+    "Contributors": [
+        "Amirreza Mirzaei"
+    ],
+    "Source": [
+        "semeval 2019 task10"
+    ],
+    "Categories": [
+        "Math Question Answering"
+    ],
+    "Definition": "You are given a geometric mathematical question. questions in this tasks often involve shapes and Geometric Relationships. you are also given 4 or 5 answer options (associated with \"A\", \"B\", \"C\", \"D\", \"E\"). Things to avoid: 1. Do not generate anything else apart from one of the following characters: 'A', 'B, 'C', 'D', 'E'. \\nEmphasis & Caution: 1. LaTeX mathematical format is used to express equations. 2. Each question is solvable with high school geometric knowledge. 3. Only give one answer for each question.",
```
danyaljj: > questions in this tasks 

capitalization 

> 1. LaTeX mathematical format

unclear

danyaljj:
Thanks for the updates. 

In my earlier point regarding "Latex", I  was **not** implying that we should define Latex. 
But rather, we should avoid introducing a new term/concept when it's unnecessary. 
Basically, let's cut to the chase, without saying "LaTeX mathematical format" or "the typesetting software known as LaTeX" which is irrelevant to the task. Just directly talk about equations/symbols/operations. 


Do you think that makes sense?

amirrezamirzaei:
Im not really sure about omitting "Latex" entirely, we need a way to express the equations and other mathematical expressions in this task. if we still use this format to show our equation and just omit "Latex" from the description there might be some confusion about what our equations mean and we cant really express the equations in any other format.

danyaljj:
All that I am saying is that: 
 - many average speakers of English know about +, /, x, etc, since we learn about them at school 
 - Latex is a technical term by a few people 
 - Wherever possible, we should stick to a terminology/framing that is understood by many people.

swarooprm (commented):
@amirrezamirzaei Could you address Daniel's comments?

danyaljj:
Upon another look, I see that there are some Latex-specific commands used in the task. For example, `\overline`, `\triangle`, `\frac{}{}`, etc (which is beyond basic algebraic equations). As such, I am fine if we move ahead with this PR as-is.

danyaljj:
Merged with the main branch as well and dropped the repeated instances.

PR #80, opened 2021-08-08
Task 122-125: New Conala tasks

kurbster:
Added more arithmetic tasks.


kurbster:
I put the source value for each task as conala. The idea for each task came from conala however, the data itself is synthetic. Should I have put synthetic instead?

swarooprm:
Conala as source is fine. Please name your files starting from 122, since task121 is already taken in the PR before yours.

danyaljj:
- task122_conala_list_index_addition: is a bit convoluted. Try it on your roommates/friends and see if they can solve it. If they can't,  that's a problem.
 - "that the same": typo? 

- task123_conala_sort_dictionary looks reasonable to me. I would include a few more negative examples. For example, one where the output is only a sorted list of "firsts", instead of the dictionaries. 

- task124_conala_pair_averages looks reasonable to me. 

- task125_conala_pair_differences looks reasonable to me.

kurbster:
For task 122 I reworded the definition it to make it more clear. I had to explain what _index_ meant to my friends, but then some were able to get it right away and more got it after seeing one example. One friend still couldn't get it (he was a communications major) then when I typed the example vertically and stacked the lists on top of each other he was able to see the pattern and figure it out.

swarooprm:
Looks good! One comment, In task122, Can you define "index". May be you can say something like this, index represents the position of an element in a list.
Everything else looks good!

danyaljj:
I still find the instructions for `task122_conala_list_index_addition` confusing. 
For example, in "You should add the values at every index for every list." 
 - "add the values" is unclear (sum the values?)
 - "list" is ambiguous (inner list or outer list?)  Might help to spell it out: "inner list"
 
It might be possible to rephrase the following to make it simpler: 
"The output should be a list where the element at each index is the sum of all lists at that index."
would it be more clear to phrase this way?  "...the sum of the elements of the inner lists at that index. "

danyaljj (approved):
Looks good to me! Thanks!

PR #81, opened 2021-08-09
Task 126-131: SCAN Dataset

eshaanpathak:
Hi! Here are the tasks for the SCAN dataset (https://github.com/brendenlake/SCAN). There are 6 tasks split up by all, short, and long lengths and command to action and action to command mappings. The all and short length tasks have 6,500 instances while the long length, command to action task has 3,920 instances and the long length, action to command task has 1,836 instances. This totals to 6500 * 4 + 3920 + 1836 = 31,756 instances. Please let me know if there are any changes I should make!


swarooprm (changes requested):
Thanks for the PR @eshaanpathak ! Good Start.
Some feedbacks:
1. Definition field looks narrow. You need to define terms like "sequence of actions". You can also say "sequence of actions such as TURN_LEFT , TURN_RIGHT etc." Another information probably you want to say is write all commands in capital. Also, may be you need to define all possible commands (if they exist) in the definition so that model knows not to predict anything beyond these. In summary, how will you instruct your friend to do this task, how much information is enough for your friend to do this task, all those should be in the definition.
2. Definition should be very specific to the task and should not be generic. Currently, there exists a significant overlap between definitions across tasks.
3. [minor] rename task files so that they are more informative about the task, currently the names look similar to each other.
4. [optional] Also dd a link to the scan dataset in the source field in parenthesis.

swarooprm:
@eshaanpathak Please share your email id with me, so that I can add you to a spreadsheet that contains various resources to find datasets for creating tasks.

eshaanpathak:
I see what you're saying and have made the changes. In the most recent commit, I avoided surrounding commands with quotations or capitalizing them if they are at the beginning of the sentence in the definition to avoid confusing the models. I just sent your gmail and AI2 email addresses an email. Please let me know if there is anything else I should edit!

EDIT: How do I run the status checks on GitHub?

eshaanpathak (commented):
Sorry, I misclicked which created this review lol.

swarooprm:
Please do "git pull" and merge with the main. This will enable automated checks to run.

swarooprm:
New definitions look much better! 
However I am not entirely sure whether we should have 3 tasks for multiple versions ("long", "short" and "all") of a single task. Their definition and instances are quite similar. Beyond this concern, it's ACCEPT from my side @danyaljj.

[Minor] suggestion: instead of starting with a non-captial letter, you can use "The" so that the sentence looks natural. E.g. " The word 'around' means" sounds better than starting with " around means".

danyaljj:
Looks good to me. Thanks!

PR #82, opened 2021-08-10
Task 132: DAIS Dataset

Mirzyaaliii:
Please review.


danyaljj:
Thanks for the PR! 

 - I am a bit fuzzy on the task. Initially, I thought this is a paraphrasing task, but now I am not sure. In other words, what does `semantically similar sentence for a given sentence` mean here? 
 - what does `same context` mean? (make makes two sentences have the `same context`?) 
 - add negative examples that do not involve grammatical issues 
 - If there are multiple valid outputs for each instance, let's add them. 
 - Please add a link to the original paper, too: https://arxiv.org/pdf/2010.02375.pdf

Mirzyaaliii:
> Thanks for the PR!
> 
> * I am a bit fuzzy on the task. Initially, I thought this is a paraphrasing task, but now I am not sure. In other words, what does `semantically similar sentence for a given sentence` mean here?
> * what does `same context` mean? (make makes two sentences have the `same context`?)
> * add negative examples that do not involve grammatical issues
> * If there are multiple valid outputs for each instance, let's add them.
> * Please add a link to the original paper, too: https://arxiv.org/pdf/2010.02375.pdf

I updated the task file, please review it again.

danyaljj:
Thanks!  
My last suggestion is to add "paraphrasing" to the task types; FYI @swarooprm

swarooprm (approved):
Have address pending reviewer comments. Looks good! Merging now.

PR #84, opened 2021-08-11
Task 137-140: Detoxifying LMs Prompt Completion Dataset

eshaanpathak:
Hi! This is our crowdsourced data from our NAACL 2021 paper “Detoxifying Language Models Risks Marginalizing Minority Voices” (https://aclanthology.org/2021.naacl-main.190/) that we have collected for which language model completions to a prompt are less toxic, more fluent, more topical, and stylistically more similar. The definitions of toxicity, fluency, topicality, and style in these files are the exact same definitions that we used for our crowdworkers. There are 4 tasks and each task has 390 instances, totaling to 1,560 instances.

### DIFF
--- a/tasks/task137_detoxifying-lms_classification_toxicity.json
+++ b/tasks/task137_detoxifying-lms_classification_toxicity.json
@@ -13,7 +13,7 @@
     "Categories": [
         "Classification"
     ],
-    "Definition": "Given a prompt and two completions from a language model, select the completion that is the least toxic with respect to continuing the prompt. Toxicity is defined as a rude, disrespectful, or unreasonable statement that is likely to make you leave a discussion. Select either Completion A or Completion B.",
+    "Definition": "Given a prompt and two completions from a language model, select the completion that is the least toxic with respect to continuing the prompt. Toxicity is defined as a rude, disrespectful, or unreasonable statement that is likely to make you leave a discussion. Generate either Completion A or Completion B.",
     "Positive Examples": [
         {
             "input": "Prompt: Im 99 percent sure it was someone being an\n\n Completion A: accomplice , or maybe just plain old con man . whatever it was , it 's over . \"\n\n Completion B: asshole, your right?) but it's an awkward subject in which if you were in the trouble of \"me\", i mean \"fucking idiot\" jrgh... i dunno...I guess you are (in a legal context) not a stupid",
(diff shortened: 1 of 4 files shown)


swarooprm (commented):
These looks good. Some comments.
- Is there a way to remove instances or examples involving inappropriate or offensive content e.g. instance 1 in task138?
- [minor] I like explanations in positive and negative examples of all tasks as they are specific to the example. Just in one place (task 139, negative examples), I thought it's too generic and can have more details. 
       "explanation": "This prompt-completion pair is less topical because there is no relevance between the prompt and completion."
       "explanation": "This prompt-completion pair is less topical because there is less relevance between the prompt and completion."
- [minor] Instead of "Select either Completion A or Completion B.", say "Generate either Completion A or Completion B". I understand that for humans who had to select an option previous prompt was appropriate, but here model needs to generate output.

@danyaljj 
Apart from the concern regarding some offensive content, It's an ACCEPT from my side once the minor issues get fixed. Also, I am not sure whether the names of all authors of the NAACL paper should be there in the contributor field.

eshaanpathak:
Thanks for the review, Swaroop! Unfortunately, there isn't an accurate method to remove toxic content. If you look at section 4 of our paper (and some other papers that we have cited, such as "The Risk of Racial Bias in Hate Speech Detection"), toxicity classifiers, as of now, suffer from annotation biases and sampling biases. Defining toxicity and agreeing on a definition can also be pretty nuanced because what one may say is toxic is what another may say is not toxic. One may look at our instances and disagree with our crowdworkers' decisions of what is toxic and what is not, or they may even disagree with the definition of toxicity that we used from Google's Perspective API.

Furthermore, all of our language model completions are from some type of detoxified language model or method, such as DAPT, PPLM, GeDi, or Filtering. You can see our paper for the results on each of these detoxified language models or methods, but they don't work as ideally as we would like.

More importantly, we have an equal amount of instances of the White-Aligned English and African American English dialects in our dataset. AAE tends to be flagged more often than WAE as toxic and are thus thrown out, which isn't fair to AAE speakers. Models should definitely be safe, but they should also be equitable, and by removing content flagged as toxic (when it's truly just nontoxic AAE), minority groups will be disadvantaged and the model will also not be as distributionally robust. Thus, this dataset tests the model's ability to understand both of these dialects and determine which completion fits better. Overall, I would love to remove toxic content from our dataset and make LMs as safe as possible, but that comes at the cost of equity and further marginalizing minority groups.

I hope that makes sense, but definitely please let me know if you need me to clarify anything!

I would prefer that we keep the names of my co-authors because we all contributed to creating and formatting this dataset, both for NAACL and for this project. This will also be the first time that our crowdsourced data will be publicly released.

I will make the minor changes you requested.

swarooprm:
Thanks for your explanation! I understand your concern. 
ACCEPT from my side @danyaljj

swarooprm (approved):

PR #88, opened 2021-08-13
Task 141-143: Odd Man Out Dataset

eshaanpathak:
Hi! This is the data from the Odd Man Dataset (https://aclanthology.org/D18-1182.pdf and https://github.com/gabrielStanovsky/odd-man-out). There are 3 tasks, where each has 6,500 instances. Please let me know if there are any changes I should make!


danyaljj:
Thanks @eshaanpathak ! 

One question: is the odd-man-out problem well-defined for the "no-category" setup? 
I don't have a good example but I guess, for some examples, the odd-man might change, for a different category.

danyaljj:
@gabrielStanovsky wondering if you have any thoughts on whether the odd-man-out task is well-defined without the category.

eshaanpathak:
In sections 3.1 and 3.2 of the paper, they state that their crowdworkers and the authors themselves had high accuracies in the no-category setup, specifically in (1) selecting the odd-man-out without being given the category and (2) to come up with categories that define words. For (2), to avoid multiple possible correct answers, I made sure to list all of the possible categories in the definition.

danyaljj:
Got it! That makes sense! It's 👍 for me.

danyaljj (approved):

gabrielStanovsky:
@danyaljj Yep! that seems like a great addition to me, thanks @eshaanpathak !

PR #89, opened 2021-08-13
Add task 144 from SubjQA

Palipoor:
This is the first task I created from SubjQA. I will create other tasks. For this task, I chose the data points where the answer and question were both subjective.


danyaljj (approved):
Looks good to me! Thanks!

swarooprm (commented):
Looks good to me! One comment:
We already have "answer generation" category, don't add a new category: "question answering" in the readme. Instead use the answer generation category in readme. In JSON, category field is empty, please add "answer generation" category there. 
Everything else looks good and I will approve once you do the above change.

Palipoor:
@swarooprm I changed the category title and added it to the task file.

Issue #28, opened 2021-07-18
Keeping track of states


danyaljj:
https://github.com/kayburns/tom-qa-dataset

eshaanpathak:
I might be able to work on this, depending on how long it may take.

Throughout the `{tom or tom_easy}/world_large_nex_1000_{noise: 0 or 10}/{task type}_{question type}_{split}_test.txt` splits, there are 52 files that I can create 52 tasks from (I would concatenate the val and test files into one which will give 13 files in each split if we also include the train file). Will this be okay to do? Each of these tasks would test a different part of the model's ability to rationalize, though 52 tasks seems a lot. Please let me know what you think.

eshaanpathak:
I can take this on. How many tasks should there be? I was thinking 4 based off of the directories, but there are different task types and question types in the directories. Please let me know what you think!

Pinging @swarooprm on this to get his thoughts too.

swarooprm:
Yes, there can be 4 tasks at least. I agree that there can be more tasks based on the question types and task types, but that might be slightly tricky to find out.
Please follow these rules whenever you plan to break a dataset into multiple tasks.
- The difference between two tasks can not be just based on the difference in the question asked. E.g. in a typical QA dataset say SQuAD there are multiple questions, but they all belong to the same task. However, it can be possible that questions are in different formats (QA, NLI or completion) or require different skills to answer (common-sense knowledge, maths knowledge etc.), in those cases, you can create multiple tasks for each format and skill.
- In order to explain two tasks to anyone say your friend, the definition has to be sufficiently distinct. E.g. question generation for a dataset is a different task than answering it.

Feel free to ask any other question you might have.

PR #91, opened 2021-08-14
Task 151-154: Theory of Mind QA Dataset

eshaanpathak:
Hi! Here is the data from the ToM QA Dataset (https://github.com/kayburns/tom-qa-dataset and https://arxiv.org/abs/1808.09352). There are 4 tasks, where each has 6,500 instances. The tasks are formulated by the four different versions of the dataset that are listed in the “Details of the Dataset” section of their repository.

I thought this dataset was particularly intriguing because it has some elements of language grounding (without the visual aspect) and tests the model’s ability to rationalize what each person in the story knows about the world and how their knowledge lines up with reality.

Please let me know if there are any changes I should make!

This pull request is related to issue #28.


swarooprm (commented):
Tasks look good! Some minor comments:
1. Since all the tasks are about finding location, may be you should name all of them as find_location. E.g. task153_tomqa_answer_generation_hard_clean.json would be task153_tomqa_find_location_hard_clean.json
2. "The question will ask about the location of an object in the story with respect to either none or one of the three scenarios." Is none of the scenarios are also a possibility or your just wanted to say "one of the three scenarios"?
3. Task 154: There is extra space here: 
explanation": "Jacob is in the same room as Ella an    d sees that Ella moved the potato from the blue_treasure_chest to the green_suitcase."
4. Definition: Task 151-153 and 152-154 have the same definition. Maybe you can add some extra information in case of hard question tasks (153 and 154). Also, I would add more details about distractor in the last line of definition in case of 152 and 154. may be you can say "distractor sentence which does not help in finding the correct answer and is meant to confuse(?) or unrelated to the question... something in that line"

All are minor comments. Overall, tasks look good.

eshaanpathak:
1. Good idea! I just made that change.
2. None of the scenarios is also a possibility since there are questions such as "Where is the eggplant really?" and "Where was the eggplant in the beginning?". I described the three scenarios to create a precise description of the types of questions there are otherwise.
3. Good catch. I didn't see that typo.
4. That makes sense. I added that in.

See the commit I just made for changes!

@danyaljj Let me know your thoughts as well!

swarooprm (commented):
Looks good! ACCEPT from my side.

danyaljj (approved):

PR #96, opened 2021-08-16
Task 156: CODAH Dataset

eshaanpathak:
Hi! Here is the data for the CODAH dataset. There is 1 task with 2,771 instances. More information can be found at https://arxiv.org/abs/1904.04365 and https://github.com/Websail-NU/CODAH. Please let me know if there are any changes that I should make!


swarooprm (commented):
Thanks! Some comments.
1. Is there any reason why you did not create multiple tasks base on multiple categories of data (Table 1 in the paper and also mentioned in readme)? Since the reasoning skills are different, I guess you can create multiple tasks. Let me know if I am wrong.
2. I think the sentence "This task is designed to test common sense and has been adversarially crafted to test if a language model is able to truly generalize rather than exploit limitations and spurious correlations." is not really helpful to a model. It's basically a motivation and machines need not be motivated to do a task, unlike humans :) Instead of that sentence, I would add some other information in the definition that can help me understand the task better e.g. some information about the reasoning skills in case you plan to split this task into multiple tasks based on the reasoning skill in Table 1 of main paper.

Rest all looks good!

eshaanpathak:
1. I didn't create multiple tasks because there are 2,086 of the 2,771 instances that are labeled as "other". Since 75% of the data is not classified, I figured it would be best to not separate the data.
2. I will make these changes and push a commit soon.

swarooprm:
1. ok sounds reasonable. For future: If the task is novel or rare in NLP datasets, we allow creating a task with as few as 100 instances.

eshaanpathak:
Yeah, some of these categories don't have at least 100 instances, but I will take that into account for future tasks.

danyaljj (approved):
Looks good to me, thanks!

PR #104, opened 2021-08-17
Task 167-169: StrategyQA dataset

yeganehkordi:
tasks based on StrategyQA dataset: https://github.com/eladsegal/strategyqa


tasks/README.md
```diff
@@ -139,3 +139,6 @@ Name | Summary | Category
 `task154_tomqa_find_location_hard_noise.json` | Given a hard story with distractor sentences, answer the question regarding the location of an object. | Answer Generation
 `task155_count_nouns_verbs.json` | Count number of nouns/verbs in the given sentence | Answer Generation
 `task156_codah_classification_adversarial.json` | Given a prompt, select the completion that is the most plausible. | Classification
+`task167_strategyqa_question_generation` | Given a term, write questions based on two or more facts | Question Generation
+`task168_strategyqa_question_generation` | Given a yes/no question, its answer, and additional information, decompose the question | Question Generation
```
danyaljj: should this task be named `question_decomposition`?

yeganehkordi: Yes, That's a better name. I fixed this.

danyaljj:
Some thoughts regarding the task definitions: 

- `task167_strategyqa_question_generation`:  (a) I think `questions that are easy for humans to answer based on two or more facts, but are challenging for modern AI systems` this is not adding any information (we will give these descriptions to computers). I would drop it. (b) There are some repeated content here that can be condensed. For example, these two sentences could be merged: `Create a question such that the answer is the one provided to you. The question should be about the given term.`.  (c) How about we split this task into two subtasks? one for question generation and the other one for fact generation. 

- `task168_strategyqa_question_decomposition`: looks good. Minor rephrasing: `answer. There are no redundant questions. ` -> `answer and there should be no redundant steps in the decomposition. 

- not sure if `task169_strategyqa_answer_generation.json` is well-defined. The outputs are evidence documents from Wikipedia which contain many details not not used in the input. I am not sure if even humans can reasonably do this. I am in favor to dropping it.

yeganehkordi:
- (a)(b) That makes sense. Sure! (c) Yes, That's a great idea. I think I can replace task 169 with the fact generation task.
- Thanks! Will do.
- I see what you mean. I'll drop it.

danyaljj:
Thanks! Please resolve the conflict so the automatic tests get triggered.

danyaljj (approved):
Thanks! Looks good to me!

yeganehkordi:
Thanks! All the changes have been applied.

swarooprm (commented):
Looks good. One comment: Task169, let's not create a new category 'fact generation'. We already have 'sentence generation' category, let's use 'sentence generation' category for this.

swarooprm:
I have changed the category in the readme, but am unable to change it in the file.

yeganehkordi:
Thanks! I'll fix it now.

PR #108, opened 2021-08-19
Add tasks 171-175

Mehrad0711:
Adding more datasets from SPL project


danyaljj:
Thanks! 

Now that we're adding more tasks for other languages, wondering if you can help us address #113, before we merge this PR.

eshaanpathak:
Can't these MT tasks also go the other way around? For example, Spanish to English too.

Mehrad0711:
Hi @danyaljj, 
Thanks for your suggestion! I added input/ output language tags for the SPL tasks.
Please let me know if there's anything else.

Mehrad0711:
> Can't these MT tasks also go the other way around? For example, Spanish to English too.

Yes. The sentences are actually aligned between languages so you can go from any source language to any target language within the 11 languages supported. There are 2 domains in SPL so there will be a total of 220 tasks. After adding all tasks with English sources, I can continue adding other directions if the main authors think this is indeed a good addition.

CC: @danyaljj, @swarooprm

danyaljj:
Looks good to me, thanks!


> After adding all tasks with English sources, I can continue adding other directions if the main authors think this is indeed a good addition. 

Sure we would be happy to see them in future PRs,!

swarooprm (commented):
Looks good!
[minor comment]: Define "special placeholders" like you have defined other terms.

swarooprm (approved):
Merging!
Looking forward to more such tasks in reverse and other languages as you have mentioned in your comment.

swarooprm:
Updated Readme to resolve merge conflicts.

PR #109, opened 2021-08-19
Task 176 and Task 184 from dataset BREAK 

Palipoor:


danyaljj:
Thanks! 

The task description looks good to me. 


 - Each input comes with a long list of "allowed tokens". I wonder if it makes sense to drop their "\n" character to make it shorter and increase the readability of the raw text inputs. 
 - Do you think there is any other task that we can define here? Here are suggestions: 
       (1) given a decomposition, verbalize it (write a natural language question corresponding to it) 
       (2) given a decomposition (without the question, follow its steps and answer it

Palipoor:
@danyaljj 

- It makes sense. I will do it.
- The task (1) you mentioned sounds good and doable, but I'm not sure about the task (2) as the questions are kind of like queries to a knowledge base and I'm not sure how they can be actually answered with a generated text.

danyaljj:
Got it, makes sense.  👍

danyaljj:
Thanks, looks good to me.

swarooprm (approved):
Looks good! Merging.

Some comments:
1. Are there any answer generation tasks you can create from the BREAK dataset?
2. There can be more tasks potentially created from the SubjQA dataset.
3. Once you complete creating tasks from a dataset, please mark the cell green in the spreadsheet.

Comments are mostly for creating new tasks. Please create new PR when you do so.

PR #111, opened 2021-08-20
Task - 178: QuaRTz

Mirzyaaliii:
Please review.


danyaljj:
- blank (Question)": fill-in-the-blank statement? 
- Include a separator between different elements of the input? For example: 
```
.... \n Option1: fall \n Option2: rise \n Qualitative Statement: 
```
 - Why the naming of "Quantitative Statement"? Alternatives could be "explanation" or "rationale".
 - Do you think there is any other task that can be constructed based on this dataset? For example, generating the explanation/quantitative statement, given the rest of the question + answer.

Mirzyaaliii:
> * blank (Question)": fill-in-the-blank statement?
> * Include a separator between different elements of the input? For example:
> 
> ```
> .... \n Option1: fall \n Option2: rise \n Qualitative Statement: 
> ```
> 
> * Why the naming of "Quantitative Statement"? Alternatives could be "explanation" or "rationale".
> * Do you think there is any other task that can be constructed based on this dataset? For example, generating the explanation/quantitative statement, given the rest of the question + answer.


All the changes are done and also created a new task from this dataset.
Please review again.

danyaljj:
Thanks! Looks good to me!

swarooprm (commented):
Looks good! Approving.
One comment you can incorporate when you create your next pull request: typically we want to have some information in the definition that is exclusive to the task. E.g., in this task, you could have a statement something like this:  "In order to answer the question, you typically need to perform a qualitative comparison based on the explanation statement". Without this kind of information, the definition looks too generic.

swarooprm (approved):

PR #114, opened 2021-08-21
Tasks 179 - 181 from the paper "Effective Crowd-Annotation of Participants, Interventions, and Outcomes..."

Palipoor:
Hi. I split their data into three tasks, but I don't know what should the category be. I put the closest thing I could though.


Palipoor:
I realized I've mistakenly named previous tasks, so I renamed those too.

tasks/README.md
```diff
@@ -101,8 +101,8 @@ Name | Summary | Category
 `task096_conala_list_index_subtraction` | Given a list of numbers subtract each element by its index in the list | Answer Generation, Arithmetic
 `task097_conala_remove_duplicates` | Given a list of numbers remove all of the duplicates in the list | Text Modification, Arithmetic
 `task098_conala_list_intersection` | Given a two lists of numbers find the intersection of the two lists | Answer Generation, Arithmetic
-`task_111_asset_sentence_simplification` | Given a sentence, simplify it so it can be understood by non-native English speakers | Generation, Paraphrasing
-`task_112_asset_simple_sentence_identification` | Given two text pieces, choose the one that is simpler and easier to understand by non-native speakers | Answer Generation, Sentence Comparison
+`task111_asset_sentence_simplification` | Given a sentence, simplify it so it can be understood by non-native English speakers | Generation, Paraphrasing
```
danyaljj, line 104: yup, this is the correct format.

tasks/task179_participant_extraction.json
```diff
@@ -0,0 +1,2577 @@
+{
+    "Definition": "In medical studies, treatments are tested within a group of study participants. You will be given a sentence of a study report in which your task is to list the tokens that  gives information about the participants of the study. If no information about the participants is mentioned, just output \"not found\". Relevant information about participants include: gender, medical conditions, location, number of people participating. Do not contain participant mentions without relevant information.",
```
danyaljj: "tokens that  gives": typo

tasks/task179_participant_extraction.json
```diff
@@ -0,0 +1,2577 @@
...
+        {
+            "input": "Non-inferiority was defined as an upper 95 % confidence limit ( CL ) for the difference in success of eradication in the amoxicillin and penicillin V treatment groups of < or=10 % .",
+            "output": [
+                "not found"
+            ]
+        },
+        {
+            "input": "In conclusion , in 60-year-old subjects antioxidant supplementation had no effect on the exercise-induced increase in oxidative stress as measured by free radical products of antipyrine .",
+            "output": [
+                "60-year-old,subjects"
```
danyaljj: why is there a comma between "subjects" and "60-year-old"?

tasks/task179_participant_extraction.json
danyaljj: - "list the tokens": if the answer is always a sub-phrase/sub-string from the given sentence, it might be easier to use any of these keywords (instead of "tokens" which is less common among lay speakers of English).

danyaljj: Combine the following two sentences? 
> Relevant information about participants include: gender, medical conditions, location, number of people participating. Do not contain participant mentions without relevant information. 

For example:  
> Mentions of participants should include relevant information such as gender, ....

tasks/task180_intervention_extraction.json
```diff
@@ -0,0 +1,2587 @@
...
+        {
+            "input": "A physiotherapy service to an emergency extended care unit does not decrease admission rates to hospital : a randomised trial .",
+            "output": [
+                "physiotherapy,service"
+            ]
+        },
+        {
+            "input": "Antioxidant supplementation and exercise-induced oxidative stress in the 60-year-old as measured by antipyrine hydroxylates .",
+            "output": [
+                "Antioxidant,supplementation"
```
danyaljj: A concern: currently, if a model outputs "supplementation,Antioxidant," (which is a correct answer) we would punish it..

tasks/task181_outcome_extraction.json
```diff
@@ -0,0 +1,2577 @@
+{
+    "Definition": "In medical studies, treatments are tested within a group of study participants. To determine if a new treatment works, various outcomes are measured in the people who take part in the study. You will be given a sentence of a study report in which your task is to list the tokens that  gives information about the outcomes of the study. If no information about the outcome is mentioned, just output \"not found\". Outcomes contain: outcomes measured in patients, like blood sugar, outcomes regarding the intervention, like effectiveness, costs, the score on a medical test or questionnaire, positive or negative events in the patient groups, like quitting smoking, or adverse reactions. Do not mention numbers or results, interpretations of outcomes, outcome mentions without relevant information.",
```
danyaljj: Same as earlier comments ("tokens that  gives " typo, etc).

tasks/task180_intervention_extraction.json
danyaljj: This concern applies to all 3 of the tasks here.

tasks/task179_participant_extraction.json
Palipoor: Umm, I don't know how to present the outputs. The annotation they have is a binary array with a 1 for each token that matches and 0 for others. So I thought of presenting them as separate tokens. I'd be glad if you have any suggestions.

tasks/task180_intervention_extraction.json
Palipoor: Yes, you are right. Should I add this to the description of tasks? That they should be in order?

tasks/task179_participant_extraction.json
danyaljj: I see these options: 
 (1) Expect only one of the responses: Basically, a system that returns one of the correct spans gets the full credit. Requires minor changes in the instructions and also having list of answers, like this: 
```
"output": [
                "60-year-old",
                 "subjects"
]
```
 (2) Include all the valid combinations. For example:  
```
 "output": [
                "60-year-old,subjects", 
                "subjects,60-year-old"
]
```
 (3) define an expected order in the definition (e.g,. in the same order which they appear in the text).

What do you think?

Palipoor: I think the third one makes more sense.

danyaljj:
Looks good to me!

swarooprm (approved):
I am not sure if participant, intervention, and outcome are standard definitions in the healthcare domain. But I think it makes sense. Merging.

PR #115, opened 2021-08-22
Task 182: DuoRC dataset

yeganehkordi:
Based on: https://github.com/duorc/duorc


danyaljj (approved):
Looks good to me!

swarooprm (approved):
Looks good to me! I am merging this task. However, note that typically we do not create tasks that ask to generate question and answer at once, instead we create 2 tasks one for question generation and another for answer generation (e.g. see task001 and task002). If you want you can create 2 additional tasks one for question generation and the other for answer generation on DuoRC without changing this task.

yeganehkordi:
Thanks, Sure! I'll create 2 additional tasks.

PR #116, opened 2021-08-22
Task 183 add rhyming task

ethankim00:


danyaljj:
Thanks! Can you check why the tests are failing? You can click on "details" above.

tasks/task183_rhyme_generation.json
```diff
@@ -0,0 +1,6032 @@
...
+    "Contributors": [
+        "Ethan Kim"
+    ],
+    "Source": [
+        "Synthetic"
+    ],
+    "Categories": [
+        "Text Generation"
+    ],
+    "Definition": "Given an input word generate a list of words that rhyme exactly with that word",
```
danyaljj: Some suggestions: 
 - Say that, if no word that rhymes found, output "No ...". 
 - One current issue is whether we have a good way of evaluating the output (like, what if instead of "pooh queue poo koo du" one generates "poo koo du pooh queue"?) To address these concerns, I suggest: 
     - Change the task description such that, it asks for one word, rather than a list of words (which can be long and open-ended).    
     - You change the output to a list. For example, 
```
"output": [
    "pooh queue poo koo du"
]
```
could be: 
```
"output": [
    "pooh", "queue", "poo", "koo", "du"
]
```

swarooprm:
@ethankim00, Seems like there is a test failure. Can you address?
File "src/test_all.py", line 45, in <module>
    assert type(i) == str, f'the output is not a string'
AssertionError: the output is not a string

swarooprm:
@ethankim00, Commenting to make sure that you have seen my earlier comment.

danyaljj:
Looks good to me!

swarooprm (approved):
Looks good to me. A minor typo "If not rhyme is found return \"No\""" should "If no rhyme is found return \"No\""". You can fix it in your next PR.

Congrats on your first task. We encourage you to create more such tasks.

PR #120, opened 2021-08-24
Task 184-190 SNLI Tasks, 197-204 MNLI tasks

aarunku5:
6 answer generation tasks, 1 classification task


swarooprm:
Task 183 is already taken, this should be named from to Task 184-190

aarunku5:
MNLI- 6 classification, 2 answer generation

swarooprm (commented):
1. task184-185, 186-188 names are same except task number, consider changing.
2. Since the new sentence 2 is formed by modifying existing sentence 2, I would make the task category "Text Modification", rather than answer generation. However, there are some instances where the output is not modified from sentence 2. In that case, we need to change category and also mention the same in the definition.

swarooprm (commented):
In case of MNLI:
1. Minor typos in readme, "gnre" should be "genre"
2. For tasks like 197 etc. you need to define various possible genres in the definition. Without that, the output space is not defined. Same is true for other tasks involving genre, you need to have some information about it in definition.
3. task199 looks pretty similar to SNLI classification; definition also pretty similar. You might consider changing few things e.g. asking to generate entailment instead of E.
4. Tasks such as 203 can also contain an additional category of 'sentence generation'
5. Task 199 and 204 again looks similar, consider changing the setup, definition etc.

swarooprm:
@aarunku5, Commenting to make sure that you have seen my earlier comment.

aarunku5:
SNLI changes are done.

MNLI- 204 looks at common genre between two sentences, while 199 is whether the sentences are a neutral pair or not, so though output format is the same, the task is different. Also in 199 I had Y/N to indicate yes or no. Have changed that to yes/no explicitly now to avoid confusion with SNLI tasks and 204.

swarooprm (approved):
LGTM! Merging.
One comment: Text modification should be added as categories in certain tasks.

PR #123, opened 2021-08-25
Tasks 205-208: New Synthetic tasks

kurbster:
Also, there is a bug in the current test_all.py script. The for loop on line 65 does not run. The line 55 `instances = enumerate(data['Instances'])` should be removed. After the first for loop runs on line 58 the `instances` variable will point at the end of the enumerate object. Thus when the second loop runs there is nothing to enumerate over. I've made a new test script that solves this error (and also runs faster) but did not include it, let me know if you would like me to add the new testing script as well.  Currently, there are 13 files that include input from a *Positive Example* in the *Instances*.


swarooprm:
Tasks till 204 are taken (https://github.com/allenai/natural-instructions-expansion/pull/120), I have renamed this PR. Please rename your files.

Re: issue in test file- Thanks for bringing this to our attention. I am not sure if this is a bug. Please add your changes in a new PR (saying fixing test file)

kurbster:
How do I open a pull request with a specific commit? Currently, when I go to open another PR it includes this PR and the commits in this PR. Also, when I push to my local branch the commit appears under this PR.

This code is a simpler version of what's going on. You can run this to confirm.

    arr = [1,2,3,4,5]
    instances = enumerate(arr)
    for idx, ar in instances:
        print(f"I am the index = {idx} I am the element {ar}")
    for idx, ar in instances:
        print(f"SECOND LOOP I am the index = {idx} I am the element {ar}")

danyaljj:
- `task205_remove_even_elements` looks good! 
- `task206_collatz_conjecture` looks good 
- `task207_max_element_lists` looks good 
- `task208_combinations_of_list`: I think we need to improve the task description: 
      - `combinations` is a loaded term (can be confused with its meaning in combinatorics). Use a different term here. 
      - `Choosing an element means that element cannot appear in the same list again.`: this is a key sentence that is not quite clear. For example, what does `same list` refer to?  It might be more clear if you phrase it something along this line: "all the generated lists must contain all the elements from the original list, except one element. Naturally, this results in `N` new lists, each with size `N-1`. "

danyaljj:
Looks good to me, thanks!

swarooprm (commented):
Looks good!
One comment: When you create synthetic tasks like remove even element or 3n+1 etc., add more variations e.g. remove odd element, 2n+1 or 4n+2 etc. You can either create more tasks or append in the same task. 
This way, model evaluation will be more reliable.

swarooprm (approved):

swarooprm:
Merging! Apply the above comments next time you create PR.

PR #124, opened 2021-08-25
Fixed unit test

kurbster:


kurbster:
This PR won't pass the new test, because there are old tasks that do not meet the requirement that, input from the positive examples is not included in, input for the instances.

src/test_all.py
```diff
@@ -52,26 +53,28 @@
...
-            for x_idx, x in instances:
-                for y_idx in range(len(data['Positive Examples'])):
-                    y = data['Positive Examples'][y_idx]
-                    if x['input'] == y['input']:
-                        raise Exception(f" * Looks like we have a same example across positive examples and instances! Drop the example from the instances. :-/ \n {x}\n {y}")
+            # Make sure there are no examples repeated across instances and positive examples
+            examples = [ex['input'] for ex in data['Positive Examples']]
+            for instance in instances:
+                if instance['input'] in examples:
+                    raise Exception(f" * Looks like we have a same example across positive examples and instances! Drop the example from the instances. :-/ \n {instance}")
```
danyaljj, line 73: this makes sense!

danyaljj:
Thanks for the PR! This is super helpful! 

Ideally, we should fix the tasks that causing the tests to fail in this PR. If you don't have the bandwidth to fix those, @swarooprm or I will try to look into them.

danyaljj:
Thanks again, @kurbster for the PR. I have dropped the overlapping instances that failed the corrected test and everything looks good now. ✅ 

FYI @swarooprm several of the tasks from the v1 dataset also have instance/positive examples overlap. See the changes in this commit: https://github.com/allenai/natural-instructions-expansion/pull/124/commits/654569e7ba133acb37754acba2d5ff415fb20323 We may wanna address these issues there as well.

PR #129, opened 2021-08-26
Task 210-212: logic2text

Mihir3009:


tasks/README.md
```diff
@@ -174,4 +174,7 @@ Name | Summary | Category
...
 `task191_hotpotqa_question_generation.json` | Given a set of context, supporting facts and an answer, generate the question asked based on them. | Question Generation
-`task192_hotpotqa_sentence_generation.json` | Given a context paragraph, question and corresponding answer, generate the supporting facts that helps in answering question. | Sentence Generation
\ No newline at end of file
+`task192_hotpotqa_sentence_generation.json` | Given a context paragraph, question and corresponding answer, generate the supporting facts that helps in answering question. | Sentence Generation
+`task210_logic2text_equation_generation.json` | Given a natural language interpretation, generate a command using logical operations | Equation Generation
+`task211_logic2text_classification.json` | Given a command and corresponding interpretation, classify whether it's right interpretation or not | Classification
+`task212_logic2text_classification.json` | Given a command (in the form of logical operators), divide in seven categories based on their action | Classification
```
danyaljj: Not sure what "divide in seven categories based on their action" means. Rephrase?

danyaljj:
Thanks! 
For the classification task, 
 - should we emphasize that it is the semantic type of the outermost operator? 
 - what do we mean by "superlative"? It might be helpful to briefly define each of the classes.

tasks/README.md
Mihir3009: I rephrased it. Please check and let me know changes.

Mihir3009:
A1: No, because it is not entirely dependent on the outermost operator.
A2: I have already included the definition for each category.

Please check and let me know. Thanks for the review.

swarooprm (commented):
Looks good to me. 
One comment: instead of equation generation, the category should be 'strtuctured text generation' since we already have that category in our repo, let's not create a new one.
Also, a question: do you mention in definition when to use parenthesis and which type of parenthesis in the logical form? May be I missed it in the definition.

danyaljj:
> A1: No, because it is not entirely dependent on the outermost operator. 

Then what is it dependent on? 

Also, is it possible that one input might belong to two semantic categories?

Mihir3009:
> Looks good to me.
> One comment: instead of equation generation, the category should be 'structured text generation' since we already have that category in our repo, let's not create a new one.
> Also, a question: do you mention in definition when to use parenthesis and which type of parenthesis in the logical form? Maybe I missed it in the definition.

Resolved this.

Mihir3009:
> > A1: No, because it is not entirely dependent on the outermost operator.
> 
> Then what is it dependent on?
> 
> Also, is it possible that one input might belong to two semantic categories?

-> According to their paper (https://arxiv.org/pdf/2004.14579.pdf), it quotes "Each description (here, command) involves exactly one type of logic".

-> They haven't specified that it only depends on the outermost operator, but you have to look whole command and judge that which category it falls into. I don't know how to clarify this like where to look. Can you please suggest something if we can include it in the definition to make it clear?

Thanks.

tasks/README.md
```diff
@@ -190,6 +190,12 @@ Name | Summary | Category
 `task170_hotpotqa_answer_generation.json` | Given a set of context and supporting facts, answer the question asked based on them. | Answer Generation
 `task191_hotpotqa_question_generation.json` | Given a set of context, supporting facts and an answer, generate the question asked based on them. | Question Generation
 `task192_hotpotqa_sentence_generation.json` | Given a context paragraph, question and corresponding answer, generate the supporting facts that helps in answering question. | Sentence Generation
+<<<<<<< HEAD
```
danyaljj: The git pull merge is not done correctly here.

danyaljj:
Got it, the paper makes it clear. Thanks! 

Note that your git merge is not done correctly (see the tasks.md file).

tasks/README.md
```diff
@@ -190,6 +190,11 @@ Name | Summary | Category
...
+`task210_logic2text_structured_text_generation.json` | Given a natural language interpretation, generate a command using logical operations | Structured Text Generation
+`task211_logic2text_classification.json` | Given a command and corresponding interpretation, classify whether it's right interpretation or not | Classification
+`task195_sentiment140_classification.json` | Given a tweet text, classify it into positive or negative | Classification
+`task196_sentiment140_answer_generation.json` | Given a tweet text and boolean question, generate answer yes or no | Answer Generation
+`task212_logic2text_classification.json` | Given a command (in the form of logical operators), classify command in one of seven logic types | Classification
 `task193_duorc_question_generation` | Writing a question based on a given plot | Question Generation
 `task194_duorc_answer_generation` | Given a plot and a question, answer the question based on the plot | Answer Generation
 `task195_sentiment140_classification.json` | Given a tweet text, classify it into positive or negative | Classification
```
danyaljj, line 195: Can you sort the lines you added such that (e.g., task212 should after 195).

swarooprm (approved):
Looks good! Merging.

PR #137, opened 2021-08-28
task 224

amirrezamirzaei:
task 224 using [scruples](https://arxiv.org/abs/2008.09094) Anecdotes dataset.


danyaljj:
Thanks! Minor suggestions below; otherwise it looks good to me. 

 - "anecdote. each anecdote " -> "anecdote, which ..." 
 - "is ethically" -> has acted ethically? 


For future PRs: we used to follow "Emphasis & Caution:" and "Things to avoid" in the past; but since a while ago, we have decided to deviate from this rule and use the natural structure/phrasing of the tasks (i.e., just state the task as-is, without adding such artificial structures).

danyaljj:
Thanks! Looks good to me!

swarooprm (approved):
Looks good to me! Merging.
Also, consider creating more tasks from the scruples dataset.

PR #140, opened 2021-08-29
task 225-226: Stackoverflow english learner tasks

Maitreyapatel:
I have created two example tasks from Stack Overflow - English Language learner category. 
These tasks are very complex in terms of open ended, length of answers etc. Therefore, I think these should be a good addition.


danyaljj:
Thanks! 

 - is there a paper on this dataset? If so, please add it to the files. 
 - For the generation task, do we know what is human performance? (can avg humans do this?

Maitreyapatel:
Please find few more details about the dataset:
1. This data isn't used in any paper as far as I can tell. I collected through Stack-Overflow archives. 
2. Because of the same reason there is no human performance. Although, the task is created keeping in mind the stack-overflow metric (i.e., upvotes) to make sure that the answer given by users are accepted by others. If upvotes are good then answer is in good quality else it's not.

I created these tasks considering the possibility of open-ended QA on the tasks which don't require too much external knowledge but it needs to know core understanding of language.    

Let me know your thoughts!

danyaljj:
> Please find few more details about the dataset:
> 
> 1. This data isn't used in any paper as far as I can tell. I collected through Stack-Overflow archives.
> 2. Because of the same reason there is no human performance. Although, the task is created keeping in mind the stack-overflow metric (i.e., upvotes) to make sure that the answer given by users are accepted by others. If upvotes are good then answer is in good quality else it's not.
> 
> I created these tasks considering the possibility of open-ended QA on the tasks which don't require too much external knowledge but it needs to know core understanding of language.
> 
> Let me know your thoughts!



Thanks! 

My concern is the quality of the input/outputs. If we can convince ourselves about the quality (of input and/or output), I would be much happier. 

 - One immediate thing that can be done is to drop redundant characters such as `\n` that reduce the readability. 
 - Can you elaborate more on "keeping in mind the stack-overflow metric (i.e., upvotes)"? What thresholds did you use and how did you apply them?

Maitreyapatel:
Yes, I also gave a thought towards the quality. 
For the first task, QA pairs are selected if and only if both has at least one upvotes and answer is accepted by the user who asked the question.
Second task is mainly towards how acceptable the answer is. So, for this task again there are two possibilities either answer having very high upvotes or the upvotes are in negative. Therefore, first I filtered the question having at least one upvotes and having at least two answers (also, one of them should be an accepted answer). Now, one answer is acceptable and another is don't.

danyaljj:
Yeah, I think we should choose these thresholds tighter so that the resulting data is very good. 
You might be worried about the dataset size (since we might be filtering many examples), but that shouldn't be important, as long as we stay north of 200 instances. Quality >> Quantity.

swarooprm (commented):
I like these tasks in general. Just make sure to include those instances with high upvotes (upvote of 1 may not be a good filter).

swarooprm (commented):
For task 226 which is a classification task, I did not find negative samples (with label 'no'). Make sure that there are some negative samples in instances.

Maitreyapatel:
Thanks for the suggestions. I added another section of stack-overflow "English" in addition to existing "English Learner". 
I kept the threshold of at least 10 for question and 10 for it's accepted answer. Now, out of ~12000 QA pairs (considering only accepted answers) this threshold filters about ~400 questions. If this threshold looks good then I can apply them and do another suggested checks.

danyaljj:
Sounds good to me! Like I suggested earlier, try to drop the redundant character too, so that the text content is more readable.

Maitreyapatel:
Sounds good. I will recreate the task files. Thanks!

swarooprm:
Thanks for doing the changes @Maitreyapatel. I like these tasks. Just one concern: task226 has most answers 'yes', we should have some 'no' answers otherwise the task instances will be skewed.

Maitreyapatel:
Yes that's little bit skewness. But I think it's doable. There are about 450-"yes" and about 200-"no" instances. If it's required, I can balance them.

danyaljj:
I also vote for a more balanced distribution. Say around 200 yes's and 200 no's?

Maitreyapatel:
Made the data set more balanced. And now there are about ~240 yes's and no's .

swarooprm (approved):
LGTM

PR #151, opened 2021-09-02
tasks 250-253: more from SPL

Mehrad0711:
Adding remaining datasets for restaurants domain (with English as the source language).


eshaanpathak:
1. What's the reasoning for the space between the word and the apostrophe and "s" in some of the examples and instances? For example, "wendy 's". I think it would be best to remove these spaces so the LMs can better understand them better with how "'s" represents belonging and/or relation and how it is not some separate word or entity.
2. Change "natural written(formal)" to "naturally written and formal". Change "formal form" to "formal language". Remove the "However," in the next sentence.
3. Add a period in the last sentence of the definition.
4. Not sure about the asterisks and caps lock for emphasis but I can leave that to @swarooprm and @danyaljj to decide.

Looks good to me otherwise! Nice work. These are all minor changes that I recommend.

eshaanpathak (changes requested):
1. What's the reasoning for the space between the word and the apostrophe and "s" in some of the examples and instances? For example, "wendy 's". I think it would be best to remove these spaces so the LMs can better understand this text better with how "'s" represents belonging and/or relation and how it is not some separate word or entity. In other words, "wendy 's" should become "wendy's".
2. Change "natural written(formal)" to "naturally written and formal". Change "formal form" to "formal language". Remove the "However," in the next sentence. Add a period in the last sentence of the definition.
3. Not sure about the asterisks and caps lock for emphasis, but I can leave that to @swarooprm and @danyaljj to decide.

Looks good to me otherwise! Nice work. These are all minor changes that I recommend.

swarooprm (commented):
Looks good! 
I was wondering if you can create more translation tasks for longer text. Here, every instance is limited to questions and contains shorter text. May be you can create new tasks in this way.

Mehrad0711:
Hi @eshaanpathak, Thanks for your suggestions!
I addressed some of your comments in 2. (since the data is originally proposed for semantic parsing, "formal form" is preferred "formal language" to avoid confusion with meaning representation used for this datasets i.e. ThingTalk)

Regarding 1. The datasets for this task are all processed with Stanford CoreNLP which splits contractions into two tokens. However, it inadvertently splits words like "wendy's" or "McDonald's" as well. For BPE based tokenizers like BERT it wouldn't matter cause they will split these tokens too but it may make a difference for SPM based tokenizers. Depending on the use cases, users may need to apply some additional post-processing: replacing placeholders with actual values, removing quotations marks around entities, detokenization as desired.

Mehrad0711:
> Looks good!
> I was wondering if you can create more translation tasks for longer text. Here, every instance is limited to questions and contains shorter text. May be you can create new tasks in this way.

Yes, I can look into other datasets for the translation task if I get the chance. Do you have any suggestions?

swarooprm (approved):
LGTM!
I don't have any particular dataset in mind, but I am sure there will be many such datasets. Thanks!

PR #155, opened 2021-09-02
Task 268: CaseHOLD

yeganehkordi:
Based on: https://arxiv.org/pdf/2104.08671.pdf


eshaanpathak (changes requested):
1. Can you elaborate more on the task in the definition? A few things to expand upon would also be to maybe elaborate on what a judicial decision is and what a holding is. Might also be best to include some background information on what this task is with all this legal talk.
2. Make the output just be "(A)" or whatever letter the correct answer is without the text that follows just to simplify things. For example, "(A) holding that possession of a pipe bomb is a crime of violence for purposes of 18 usc 3142f1" would just be "(A)".
3. Maybe add a new line character ("\n") or even two ("\n\n") right before it says "Holding statements:" in each example just to separate the decision from the holdings so they can be seen easily.

Overall, this is an interesting task with some recommendations I have. Nice job.

slowy07 (approved):
LGTM on ``README.md`` no spelling mistakes

yeganehkordi:
@eshaanpathak @slowy07 Thanks for the reviews! I updated the task.

swarooprm (approved):
I like the diversity this task is adding, we did not have such judicial-related tasks. Keep it up!
2 suggestions [optional]:
1. You can create an incorrect answer generation task
2. May be you can introduce a new category here, since the domain here is slightly different
3. More tasks in this domain is highly encouraged since we don't have such type of tasks in our repo.

Issue #131, opened 2021-08-27
Counterfactual reasoning task 


danyaljj:
https://www.semanticscholar.org/paper/Counterfactual-Story-Reasoning-and-Generation-Qin-Bosselut/4d16457cded23bce6eaa91cd17aefd22af2279f0

Maitreyapatel:
This sounds interesting problem. Let me create few tasks from this data.

swarooprm:
Dataset here is the "Time Travel" dataset, we have task065 and task066 which used part of the original crowdsourcing template, but more tasks can certainly be created.

PR #156, opened 2021-09-02
Task 269-270: Counterfactual story generation

Maitreyapatel:
Submitting the two tasks from issue #131 . 
First, task focuses on generating new story ending based on counterfactual context.
Second, focuses on generating counterfactual context based on new and old endings such that it is counterfactual to initial context and justify the differences.


tasks/README.md
```diff
@@ -209,4 +209,6 @@ Name | Summary | Category
 `task227_ClariQ_classification`	| Given a query and its clarification, classify whether clarification is proper or not by providing 'Yes' or 'No' | Classification
 `task243_count_elements_in_set_intersection.json` | Count number of elements in the intersection of two given sets | Counting
 `task244_count_elements_in_set_union.json` |  Count number of elements in the union of two given sets | Counting
-`task245_check_presence_in_set_intersection.json` | Check presence of an element in the intersection of two given sets | Answer Generation
\ No newline at end of file
+`task245_check_presence_in_set_intersection.json` | Check presence of an element in the intersection of two given sets | Answer Generation
+`task269_CSRG_Counterfactual_story_Generation.json` | Given premise, initial context with ending, and counterfactul context, generate new story ending supporting counterfactual. | Answer generation
```
yeganehkordi: There was a typo in definition, could change to: Given the premise, initial context with ending, and counterfactual context, generate a new story ending supporting counterfactual.
And same here: Given the premise, initial context with ending, and counterfactual context, generate a new story ending supporting counterfactual.

yeganehkordi (commented):
Thanks!
There are some typos in task definitions. 
- I think you can replace the definition of task 269 with: Given the premise, initial context, counterfactual context, and original ending, generate the new ending by considering the new counterfactual context in the original ending instead of the original context. Edit the original ending as little as possible to change the new counterfactual context. If necessary, counterfactual context is very different, then heavy editing is suggested. Here, Counterfactuals are conditional sentences that discuss what would have been true under different circumstances.

- And definition of task 270 with: Given the premise, initial context, original ending, and new ending generate the context involved in the new ending, counterfactual to the initial context. Here, Counterfactuals are conditional sentences which discuss what would have been true under different circumstances.

- I'm not sure, but I think text generation would be a better category for these tasks.
- Please fix the typo of "initial" and "counterfactual" in examples.
- I think you could replace "Generated context is in align... " in the examples with "Generated context is in alignment ...".

Maitreyapatel:
Thanks for the important suggestions.
I've applied them and updated the task files.

danyaljj:
Thanks! 
I found the task definition difficult to follow, particularly for `task270_CSRG_Counterfactual_context_generation`.  
Here is a suggestion: Ask your roommate/friend to read the definitions and see if they would be able to understand and solve the task without any additional help.

Maitreyapatel:
Thanks for sharing your concern. I was having the same doubt with new suggested definitions. Please refer to the new definitions. 

For task 269: The premise, initial context, counterfactual context, and original ending are given as input. Using these inputs, generate the new ending by modifying original ending. This new ending should be aligned with counterfactual context. Edit the original ending as little as possible to support the counterfactual context. If counterfactual context is very different, then heavy editing is suggested. Here, Counterfactuals are conditional sentences that discuss what would have been true under different circumstances.

For task 270: The premise, initial context, original ending, and new ending are given as input. Using these inputs, generate the new context involved in the new ending, which is counterfactual to the initial context. And this new context should explain the differences between original and new ending. Here, Counterfactuals are conditional sentences which discuss what would have been true under different circumstances.

danyaljj:
@swarooprm do we have the crowdsourcing templates for this dataset? (https://arxiv.org/abs/1909.04076) Wondering how they have defined in their crowdsourcing templates.

swarooprm:
> @swarooprm do we have the crowdsourcing templates for this dataset? (https://arxiv.org/abs/1909.04076) Wondering how they have defined in their crowdsourcing templates.

yes, we have the templates, this is also registered with the name 'time travel' in the spreadsheet, the templates are available in the google bucket shared with contributors. We already have task065, task066 from this dataset but more certainly be created.

swarooprm:
I like the tasks in this PR even though we have task065 and task066 on timetravel because:
1. These tasks are different from the classification tasks present in task065 and 066
2. The language here in defintion is also significantly different as crowdsourcing template is not used

Overall, its ACCEPT from my side.

danyaljj:
Thanks! 

My main concern is the clarity of the current definitions: 

> The premise, initial context, original ending, and new ending are given as input. Using these inputs, generate the new context involved in the new ending, which is counterfactual to the initial context. And this new context should explain the differences between original and new ending. Here, Counterfactuals are conditional sentences which discuss what would have been true under different circumstances. 


There are a lot of keywords here that make it tricky: "initial context", "original ending", "new ending", "counterfactual" and "new context". I feel that the task will be unclear unless we define these one at a time and build up towards the goal of the task. 
 - It might be worth asking your friend/roommate to read these instructions to see if they would follow them or not (and finding what are the blind spots). 
 - Would also be helpful to borrow definitions from the original crowdsourcing templates (see Swaroop's reply earlier).

swarooprm:
@Maitreyapatel, can you address Daniel's comments?

tasks/README.md
```diff
@@ -209,4 +209,6 @@ Name | Summary | Category
...
+`task270_CSRG_Counterfactual_context_generation.json` | Given premise, initial context with ending, and new counterfactul ending, generate counterfactual context which supports the new story ending. | Answer Generation
```

Maitreyapatel:
@danyaljj Can you please review following new description for example before I push the changes ?


Task269 OLD description:  The premise, initial context, counterfactual context, and original ending are given as input. Using these inputs, generate the new ending by modifying original ending. This new ending should be aligned with counterfactual context. Edit the original ending as little as possible to support the counterfactual context. If counterfactual context is very different, then heavy editing is suggested. Here, Counterfactuals are conditional sentences that discuss what would have been true under different circumstances.

Task269 NEW description: A \"premise\" is a statement that an argument claims will induce or justify a conclusion. \"Initial Context \" is the original story plot having specific \"original ending\" of the story. Moreover, the \"counterfactuals\" are conditional sentences that discuss what would have been true under different circumstances. Therefore, \"Counterfactual Context\" is the slight modification to the original story plot. Now, given these information as input, generate the new ending. This new ending should be aligned with counterfactual context. Edit the original ending as little as possible to support the counterfactual context. If counterfactual context is very different, then heavy editing is suggested.

Maitreyapatel:
I have changed the definitions according to above comment and renamed the files.
Please review it again.

tasks/README.md
Maitreyapatel: Renamed the files. Thanks!

danyaljj:
Thanks (and sorry for the delay). 

I think the definitions have improved, but I feel that they're not quite there yet. 
@kurbster @eshaanpathak wondering if you can take a look at this PR? The task is a bit tricky and wondering if you can provide suggestions to improve the wording.

kurbster:
I can take a look at it.

danyaljj:
@qkaren wondering if you can also share your thoughts if you get a chance.

kurbster:
For task 296, from my understanding, the _Counterfactual Context_ is a modification of the _Initial Context_. Then based on this new context modify the _Original Ending_ to create a new ending. The definition of _premise_ is not clear. I assume that _premise_ occurs before both contexts.
`premise + init context => orig ending`
`premise + counter context => new ending`
The definition could be something like this
> A premise is a statement that induces or justifies a conclusion. The initial context is the original story plot which has a specific original ending. The counterfactual context is a slight modification to the initial context. Given that the premise is the same for both contexts, generate a new ending. The new ending should support the counterfactual context.

Also, I'm confused about the first positive example. If the counterfactual context was `she keeps herself steady with her feet`. Why wasn't the original ending changed to `Suddenly, her kick sent her backward but she steadied herself. The chair was stable and didn't fall. Susie didn't hurt her head and was fine.` The fourth positive example is a similar example where the new ending is a negation of the original, i.e `the baby did not get piercings`. 

I'm also confused by the first negative example, would the correct ending be `... she felt sick. She didn't cancel but no one showed up.` Or `... When she woke up Saturday morning no one showed up.` The counter context only said no friends showed up. So would she still be sick or not? Maybe no one showed up because she was sick.

For task 270, I think the definition is better because it explains that _premise_ is the same for both contexts. I would still make it more concise `A  premise is a statement that induces or justifies a conclusion.`

swarooprm:
@Maitreyapatel,  can you address reviewer comments?

Maitreyapatel:
Sorry for the delay...! 

Thank you for the suggestions. Please find below comments:

* Your doubt on generated new endings based on counterfactual context does make sense. However, there can be many possible ways to define the new ending. Here, the authors created this dataset with constraints that minimum modifications should be done on the original ending. Now, in your example the new context is counterfactual to the initial context. And this new context may / may not suggest the changes in new ending from original ending. So, counterfactual context is independent of original ending. For example, `she keeps herself steady with her feet` does not imply that she does not have to fall in the ending. But this is just modification of initial context `She kept kicking the counter with her feet.`. Now, the new generated ending says the even though she was steady, she fell because of the earthquake.
* For the first negative example: again there can be many possible different endings. Therefore, your answers are correct but negative generated ending does not consider this required change. Hence, it is wrong.
* I will make sure to update the definitions based on your suggestions.

@kurbster I hope this answers your questions. If so, I'll update the PR after your approval.

swarooprm:
@kurbster, check and let the author know your thoughts?

qkaren:
Thank you for the discussion! The task definitions are correct. I'm just thinking that we could possibly simplify them further (e.g., we don't have to involve some of the concepts like "premise" etc, cuz they are not very crucial). E.g., for Task269, we can probably say sthm like:
"
Given an initial story context, original story ending, and a counterfactual context, the task is to generate a new story ending that is aligned with the counterfactual context and is as close to the original ending as possible.

More specifically, a story context contains two sentences. A counterfactual context is a slight modification to the second sentence in the initial context (while keeping the first sentence the same). A story ending contains three sentences. A new story ending should edit the original story ending as little as possible to regain coherence with the counterfactual context.
"

kurbster:
@Maitreyapatel thank you for your comments the examples are clear now! Also, I agree with @qkaren her definition is concise.

Maitreyapatel:
@qkaren @kurbster Thank you for the suggestions. I will update the definitions and commit the changes.

Maitreyapatel:
@swarooprm I have updated the definitions as everyone suggested. 
Please review them and let me know if we need to modify it even further.

swarooprm (approved):
LGTM!

Issue #113, opened 2021-08-20
Language indicator key/values to the schema 


danyaljj:
Add language indicator to the schema: 
 - "Input_language": language of the input/instructions 
 - "Output_language": language of the responses/output 

We can specify these languages via their [two-digit ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).

danyaljj:
Assuming that "Input_language" refers to the input instances, we should also have an "instruction_language".

PR #157, opened 2021-09-02
Tasks 271-273: Europarl

Mihir3009:
2 Sentence Generation Tasks and 1 Classification Task. Please review.


yeganehkordi (commented):
Thanks! Looks good to me. 
We already have the "Translation" category in the tasks. I think this would be a better category for the first two tasks.

Mihir3009:
> "Translation" category

I updated the "Translation" category for the first two tasks. Thanks.

danyaljj:
Please use the language indicators as suggested in #113

Mihir3009:
> Please use the language indicators as suggested in #113

Resolved and files are updated.

swarooprm (approved):
Looks good!
We have relatively fewer multi-lingual tasks, so more tasks are encouraged.
Make sure that you and @Mehrad0711 don't end up creating the same task.

PR #161, opened 2021-09-02
Task 281 Points of Correspondence detection

Palipoor:


Maitreyapatel (commented):
Looks good to me. But there is an issue with output of second negative example. "sentence 2" is missing and "sentence" is missing from "sentence 3". 
Also, can you add the link to the paper or dataset ?

Palipoor:
@Maitreyapatel   thanks. Fixed them. The missing "sentence 2" was intentional.

Maitreyapatel (commented):
It make sense. Thanks for the update!

swarooprm (approved):
I like this task. This adds to task diversity.
We might need to keep it in training spilt though since annotations are not always reliable.

swarooprm:
Also see if you can create more tasks from the same data. Say by dividing it based on place, person or thing (if those data is already separated) or by asking to generate the type of entity etc.

Palipoor:
@swarooprm Sure! thanks.

PR #166, opened 2021-09-03
Task 287: CaseHOLD

yeganehkordi:
Incorrect answer generation task based on CaseHOLD dataset
Also, I changed the category of CaseHOLD and overruling tasks to legal.


swarooprm (commented):
1. You don't need to say this: "The four incorrect answers are other holding statements." It can confuse the reader/model because ultimately you want model to generate single incorrect answer.
2. Add incorrect answer generation along with legal in the categories.
Everything else looks good. Just verify that instances contain incorrect annotations and not the correct annotation.

yeganehkordi:
Thanks! I'll fix them now.
Should I add incorrect answer generation in the README file as well?

swarooprm:
> Thanks! I'll fix them now.
> Should I add incorrect answer generation in the README file as well?

Yes add in the readme as well. Thanks!

yeganehkordi:
Thanks, Done.

swarooprm (approved):
LGTM!

PR #170, opened 2021-09-03
Task 292-294: StoryCommonsense

Mirzyaaliii:
task292_StoryCommonsense_character_text_generation  -->  Given a short story, provide all the characters in that story.
task293_StoryCommonsense_emotion_text_generation  -->  Given a sentence, context (optional) and character; provide emotions expressed by the character in the sentence.
task294_StoryCommonsense_motiv_text_generation  -->  Given a sentence, context (optional) and character; provide motivation of the character in the sentence.

Please review.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -364,6 +364,15 @@ Name | Summary | Category
 `task373_synthetic_round_tens_place` | Given a list of integers round them to the tens place | Answer Generation, Arithemetic
 `task374_synthetic_pos_or_neg_calculation`| Given a list of integers multiply the negative integers by *-3* multiply the even integers by *2* | Answer Generation, Arithemetic 
 `task375_classify_type_of_sentence_in_debate` | Given a debate topic and a sentence from the debate, classify the type of the sentence. | Classification
+`task380_boolq_yes_no_question` | Given a passage and a yes/no question, answer the question based on the passage | Answer Generation
+`task381_boolq_question_generation` | Given a passage, generate a yes/no question that can be answered based on the passage | Question Generation
+`task382_hybridqa_answer_generation` | Given a question about part-of-speech tag of a word in the question, answer the question | Answer Generation
+`task397_semeval_2018_task1_tweet_anger_detection` | Given a tweet judge whether the author was angry or not | Classification
+`task398_semeval_2018_task1_tweet_joy_detection` | Given a tweet judge whether the author was happy or not | Classification
+`task399_semeval_2018_task1_tweet_sadness_detection` | Given a tweet judge whether the author was sad or not | Classification
+`task394_persianqa_question_generation` | Given a passage, generate a question based on it | Question Generation
+`task395_persianqa_answer_generation` | Given a passage and a question, answer the question based on the passage | Answer Generation
+`task396_persianqa_classification` | Given a passage and a question, check whether the question is answerable based on the passage or not | Classification
 `task376_reverse_order_of_words` |  Reverse the order of words in the given sentence | Answer Generation
 `task377_remove_words_of_given_length` | Remove all words of a given length in the sentence | Answer Generation
 `task378_reverse_words_of_given_length` | Reverse all words of a given length in the sentence | Answer Generation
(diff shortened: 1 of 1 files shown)


Palipoor (commented):
Hi,
In task 293: 
Please change `all the emotions which is ` to `all emotions which are`, or just `all emotions`.
Also, ` in the given Sentence` can be removed.
Another issue: the examples, the emotions are comma separated, which is not the case in actual instances. If the task is to list all emotions, the instances should do the same, but I think it's easier if you change the task to "one possible emotion" and change the example outputs. Otherwise, you should contain all possible permutations of, say, 4 emotions in each instance in possible outputs. Pay attention, that the outputs of instances are lists, not because the machine is going to output a list, but because each of them are possible acceptable outputs. This is also the case for task 292.

swarooprm:
Can you address reviwer comments @Mirzyaaliii ? Overall, these are good tasks as they bring new variety to our collection.

Mirzyaaliii:
> Hi,
> In task 293:
> Please change `all the emotions which is ` to `all emotions which are`, or just `all emotions`.
> Also, ` in the given Sentence` can be removed.
> Another issue: the examples, the emotions are comma separated, which is not the case in actual instances. If the task is to list all emotions, the instances should do the same, but I think it's easier if you change the task to "one possible emotion" and change the example outputs. Otherwise, you should contain all possible permutations of, say, 4 emotions in each instance in possible outputs. Pay attention, that the outputs of instances are lists, not because the machine is going to output a list, but because each of them are possible acceptable outputs. This is also the case for task 292.

Thanks for the comments!

- The First 2 changes are done.

- According to me, keeping only 1 emotion is not a good idea because sometimes for 1 thing we can have multiple emotions.
E.g., in the 2nd positive example,
Valerie can feel nervous as well as excited about her performance at the same time.
So, according to me, by providing all the emotions, machine can learn each and every emotion which are related to that event.

- Also, in the positive and negative examples, we need to write string only (as per the instruction) and in the instance, the output should be list. That is why emotions are comma-separated in the examples. (The same thing applies to task 292.)

tasks/README.md
```diff
@@ -269,19 +269,24 @@ Name | Summary | Category
...
-`task287_casehold_legal_incorrect_answer_generation` | Given a prompt from a judicial decision and multiple potential holdings, choose one of the incorrect options | Legal , Incorrect Answer Generation
+`task287_casehold_legal_incorrect_answer_generation` | Given a prompt from a judicial decision and multiple potential holdings, choose one of the incorrect options | Legal, Incorrect Answer Generation
+`task288_gigaword_summarization.json` | Given a text of article, generate headline (title) for the article | Summarization
+`task289_gigaword_summarization.json` | Given the text of an article and its title, decide whether the title is appropriate for the article | Summarization
 `task290_TellMeWhy_question_answerability` | Given a short story and a question, decide whether or not the question is answerable | Classification
 `task291_semeval_2020_task4_Commonsense_Validation` | Given two statements, choose the one that makes less sense. | Answer Generation
+`task292_StoryCommonsense_character_text_generation` | Given a short story, provide all the characters in that story. | Text Generation
```

Mirzyaaliii: Done.

eshaanpathak:
If you look at the "Files changed" tab, it seems like there are a lot of changes to the README outside of tasks 292 to 294. I would recommend copying the current README from `allenai:master` and just adding in the rows for tasks 292 to 294 to not make any other changes. Otherwise this would undo a lot of my edits in #224.

danyaljj:
@Palipoor could you check if you're happy with the updated PR?

Palipoor:
> According to me, keeping only 1 emotion is not a good idea because sometimes for 1 thing we can have multiple emotions.
Ok, so I think it's better if you change the task description to find one of the emotions(in the same manner: one of the characters) as each of the acceptable outputs in instances are one emotion/character.

Mirzyaaliii:
> > According to me, keeping only 1 emotion is not a good idea because sometimes for 1 thing we can have multiple emotions.
> > Ok, so I think it's better if you change the task description to find one of the emotions(in the same manner: one of the characters) as each of the acceptable outputs in instances are one emotion/character.

Thank you very much for your valuable suggestion and for clarifying the problems with my tasks output.
Please review again. I made the changes.

swarooprm:
Thanks @Mirzyaaliii. Let's wait for @Palipoor to see if she is happy or has any other suggestion.

swarooprm:
@Palipoor, check if you have any other feedback?

Palipoor:
Sorry I didn't see this sooner! Looks good to me!

swarooprm (approved):
LGTM!

swarooprm (commented):
Can you fix the issues in the readme?
It's showing you are changing a lot of things which are not part of this PR. Please verify that it's not changing anything else apart from these files.
I will merge once this gets fixed.

danyaljj:
> Can you fix the issues in the readme?
> It's showing you are changing a lot of things which are not part of this PR. Please verify that it's not changing anything else apart from these files.
> I will merge once this gets fixed.

I revised the README.

PR #172, opened 2021-09-03
Task 301-303: ReCoRD

yeganehkordi:
Based on: https://arxiv.org/pdf/1810.12885.pdf


Palipoor:
Looks good to me!

yeganehkordi:
Thanks!

swarooprm (approved):
1. "Your question should be answerable based on the conversation and only have one correct answer." I am not sure what do you mean by conversation here (because based on the definition I was expecting a dialogue but could not find it), do you want to replace conversation by passage/paragraph
2. The answer generation task is good, but its mostly a classification task. You can have another task to generate answer (here answer is (A), (B) etc. ,so classification, there the answer will be pure text generation). For example, look at task004 or task002.

Everything else looks good. I am approving this PR. The above changes are minor and you can fix in the next PR.

yeganehkordi:
Thanks! 
I'll fix them with the answer generation task.

PR #175, opened 2021-09-04
Task 305-308: Jeopardy

eshaanpathak:
Task 305: Normal Jeopardy Rounds (easiest difficulty with 6,500 instances)
Task 306: Double Jeopardy Rounds (medium difficulty with 6,500 instances)
Task 307: Final Jeopardy Rounds (hardest difficulty with 3,567 instances)
Task 308: Mixed Jeopardy Rounds (varying difficulty with 6,500 instances)

4 tasks which total to 23,067 instances.

If you don't know what Jeopardy is, see [this Wikipedia link](https://en.wikipedia.org/wiki/Jeopardy!) for more information.


danyaljj:
@Mihir3009 @aarunku5 wondering if you can help with the review of this PR?

Mihir3009:
Yes, I will review this PR.

aarunku5:
I am assuming the difficulty level been demarcated based on whether the questions are from the Jeopardy, Double Jeopardy, and Final Jeopardy Rounds? In which case, the difficulty level would basically be determined by how 'obscure' the trivia fact is to a human. I'm not very clear on whether a model would truly find the questions equally hard in that case. For instance:

{
input: Category: TRACK\n\nClue: On August 16, 1995 this country's Haile Gebrselassie set the world record in the 5,000-meter run at 12:44:39
output: Ethiopia
}

and

{
input: Category: ESPN's TOP 10 ALL-TIME ATHLETES\n\nClue: No. 8: 30 steals for the Birmingham Barons; 2,306 steals for the Bulls
output: Michael Jordan
}

These are both in the easy category, but a model might not find them equally easy to solve due to the way the clue is worded.

The positive and negative examples are also the same across all tasks. Can you change them to be reflective of what you mean by 'difficulty level' to justify splitting the dataset into different tasks?

eshaanpathak:
@aarunku5 Thanks for the review!

Yeah, generally what's difficult for a human isn't also always what is difficult for a language model and vice versa. That's what these tasks are meant to test, not just the general knowledge of trivia but also all other facets of trivia. We shouldn't restrict these tasks by how they are worded, i.e. the other facets, since they also test robustness to different (and odd) ways to word something. Overall, we want to see how a language model performs on them compared to human performance, as seen with [this article on Jeopardy question difficulty](https://medium.com/@pollockcolin/jeopardy-question-difficulty-1bba47770bc6).

This is just a really interesting set of tasks because trivia falls in the realm of open-domain question answering, so it's worth seeing how much general knowledge a large language model has about the world without any finetuning and how it is able to parse some weirdly formatted questions compared to others. This is something that we achieved this year with training our crossword puzzle solving AI system (see [this article](https://engineering.berkeley.edu/news/2021/05/the-language-of-probabilities/) and [this article](https://www.wired.com/story/crossword-ai-humans-way-with-words/) for more) since crossword puzzle clues can not only be general knowledge facts but also involve a lot of wordplay, puns, themes, etc. which makes them more nuanced similar to trivia.

I'll go ahead and edit the positive and negative examples later today. Thanks for catching that.

eshaanpathak:
@aarunku5 @danyaljj @swarooprm Please see my last comment and most recent commit for my changes.

aarunku5:
Thanks! @danyaljj  the tasks all look good to me.

swarooprm (approved):
LGTM

PR #179, opened 2021-09-04
Task 318-321: StereoSet

eshaanpathak:
I know this dataset is currently being reviewed in #159, but it wasn't reserved by anybody in the Excel spreadsheet or completed in the `tasks/` directory when I checked yesterday. Compared to #159, my tasks are different since I split my tasks by domain (i.e. gender, profession, race, religion), so we can evaluate how biased a language model is for each of these 4 different domains and how the results compare with each other. Such can be seen in [Section 5 and Figure 2 of this paper](https://arxiv.org/abs/1909.01326) with how different societal groups and domains experience a varying level of biases from language models.

I would appreciate if these tasks could still be considered since this is a good dataset, my tasks are different, and I spent ~5 hours creating these tasks. Thanks!

Tasks 318, 319, 320, and 321 have 723, 2,478, 2,925, and 228 instances, respectively. This totals to 6,354 instances.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -271,7 +271,7 @@ Name | Summary | Category
 `task287_casehold_legal_incorrect_answer_generation` | Given a prompt from a judicial decision and multiple potential holdings, choose one of the incorrect options | Legal , Incorrect Answer Generation
 `task290_TELLMYWHY_question_answerability` | Given a story and a question, decide whether or not the question is answerable | Classification
 `task301_record_question_generation` | Given a passage, generate a fill-in-the-gap question based on it | Question Generation
-`task302_record_answer_generation` | Given a passage and a question, answer the question based on the passage | Answer Generation
+`task302_record_classification` | Given a passage and a question, classify the answer to the question based on the options | Classification
 `task303_record_incorrect_answer_generation` | Given a passage and a question, write an incorrect answer for the question | Incorrect Answer Generation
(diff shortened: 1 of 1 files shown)


XudongOliverShen:
oops, I am open to merging if reviewers agree! where is the Excel spreadsheet btw?

swarooprm:
> oops, I am open to merging if reviewers agree! where is the Excel spreadsheet btw?

For your future PRs, please use this link to update the dataset on which you want to create tasks.
https://docs.google.com/spreadsheets/d/1Wd4fP-OVY7hVEIo8oMbzjVr4Sfcl2Hdj3hhEP_BKoWI/edit?usp=sharing

swarooprm:
@eshaanpathak I see that you have modified your definition of stereotype and anti-steoretype and made it same as Oliver's, but I would prefer those to be different and I think more different definitions and other fields are, more likely we can accept both tasks. Here's why:
When we build a demo for people, their definitions will have a significant variation among each other in terms of length, style of writing, vocabulary, defining terms etc. even for the simplest of tasks say SQuAD. We want our model to be robust against all those, so it's better if we have different variations of the same task.

eshaanpathak:
I see. I just made the changes in the most recent commit.

danyaljj:
Thanks! The tasks look good! 
Has some overlaps with #159, but not quite overlapping. Therefore, merging this PR!

Issue #174, opened 2021-09-03
Corefence datasets 


danyaljj:
Suggestion: add tasks related to coreference resolution 

 - [x] GAP: https://github.com/google-research-datasets/gap-coreference
 - [ ] CoNLL 2012 http://nlpprogress.com/english/coreference_resolution.html
 - [ ] OntoNotes  https://paperswithcode.com/sota/coreference-resolution-on-ontonotes 
 - [ ] PreCo https://preschool-lab.github.io/PreCo/

eshaanpathak:
I can do PreCo once I get the data in my email.

yeganehkordi:
I'll work on GAP.

ashok-arjun:
@danyaljj Is this issue still open? 

Can I work on the other 3, as 3 separate tasks?

danyaljj:
GAP was addressed in  #182. But I think the rest of the items are open.

PR #182, opened 2021-09-04
Task 329-331: GAP

yeganehkordi:
Based on: https://arxiv.org/pdf/1810.05201.pdf   #174


swarooprm:
@Mihir3009, can you review these tasks?

swarooprm:
@Mehrad0711, wondering if you can review these tasks?

Mehrad0711:
Thanks for your work!

In all three tasks I think it would be better to replace "\n" with a non-escaping character to avoid unwanted line splits when reading the file. Perhaps a special token like `<sep>` which can be added to the tokenizer special vocab list.

Was wondering what the motivation is behind `gap_incorrect_answer_generation` task? Is there a specific use-case for a model trained to generate wrong answers?

In the classification task, I noticed 4 labels are introduced but "Both" and "Neither" are not used in any other instances besides positive and negative examples. Maybe some examples can be added for those. Also, wondering if "Both" makes sense at all? The examples are designed not to have overlapping references afaiu.

Also please add 1 more example for positive and 1 more example for negative instances. The tests are not catching this right now (PR #216).

It looks good to me otherwise.

yeganehkordi:
Thanks for your comments!
I updated the tasks based on the comments.
For the incorrect answer generation task, I followed the task types that were created by Daniel and Swaroop. I'm not sure about the motivation, but I think some questions can be easily answered by the system without actually understanding the concept. If a system can pass both answer generation and wrong answer generation tasks with high precision, we can have a better evaluation of its understanding.

swarooprm:
Thanks @yeganehkordi.
@Mehrad0711, please let us know if you are happy with the PR or have any suggestions?

Mehrad0711:
Yes, this looks good to me now. Thanks.

PR #188, opened 2021-09-05
Task 344-347 and Task 382: HybridQA

yeganehkordi:
Based on: https://arxiv.org/pdf/2004.07347.pdf
I wasn't sure if I could add another answer generation task like task 347, which asks about the part-of-speech tag of one word in the question.

I recognized I hadn't shuffled dream and duorc datasets before; I fixed them in this PR.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -282,6 +282,8 @@ Name | Summary | Category
 `task345_hybridqa_answer_generation` | Given a question, write the part-of-speech tag for each word in the question | Answer Generation
 `task346_hybridqa_classification` | Given a question, a word, and a POS tag, determine whether the POS tag is True or False based on the part-of-speech tag of the given word in the question | Classification
 `task347_hybridqa_incorrect_answer_generation` | Given a question about part-of-speech tag of a word in the question, write an implausible POS tag to the question | Incorrect Answer Generation
+`task348_squad2.0_unanswerable_question_generation` | Given a passage, generate a question that cannot be answered based on the passage | Unanswerable Question Generation
+`task349_squad2.0_answerable_unanswerable_question_classification` | Given a passage and a question, classify whether or not the question is answerable from the passage | Classification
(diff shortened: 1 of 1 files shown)


swarooprm:
@Mirzyaaliii, can you review these tasks?

Mirzyaaliii:
> Based on: https://arxiv.org/pdf/2004.07347.pdf
> I wasn't sure if I could add another answer generation task like task 347, which asks about the part-of-speech tag of one word in the question.
> 
> I recognized I hadn't shuffled dream and duorc datasets before; I fixed them in this PR.


General comments for all 4 tasks

-> Please add more negative examples which help to understand which kind of output is not allowed for that particular task.
-> With research paper also include a dataset link in the source (see task 164).

- Task 344: 
-> Can you elaborate more on the definition??

- Task 345:
-> Looks good to me just add negative examples.
-> In the negative examples you can add something like this - the input sentence has 5 words but output generates POS for fewer words than in input.

- Task 346:
-> Please modify the description of this task (in README), it sounds a little confusing.

- Task 347:
-> In this task instead of question in the input, you can just simply form your input like this -
`
"input": write_your_input_question \n Word: word_from_question_for_which_you_want_POS,
"output": POS_of_the_given_word
`
The above form of input is a bit easier to understand or you can just simply change your input like this -
`
"input": "What is the part-of-speech tag of the word \"the\" in the following question: Who were the builders of the mosque in Herat with fire temples ?",
"output": "IN"
`
-> Also, as suggested by you, you can create a new task that generates the right POS of the given word in output (which is the same as this task except for the right answer).

yeganehkordi:
Thanks for the comments. I updated the tasks.

danyaljj:
Thanks! The PR makes changes to DuoRC-related tasks. Are these intentional?

yeganehkordi:
Thanks! Yes, I hadn't shuffled dream and duorc instances before; I fixed them in this PR.

Mirzyaaliii:
> Thanks for the comments. I updated the tasks.

Looks good to me, @swarooprm do a final check and merge.

PR #193, opened 2021-09-06
Task 353-359: CaSiNo Dataset

kushalchawla:
Request to add sub-tasks from our [negotiation dialogue dataset](https://aclanthology.org/2021.naacl-main.254.pdf).


yeganehkordi (commented):
Thanks! Some suggestions:
- Some of the instances don't have context. Could you remove them or add some explanation about them in the definition? 
- It's better to separate context and utterance with a "\n".
- Task 355: I'm a bit confused, other-need doesn't look like selfish.
  > "other-need is a selfish negotiation strategy."

- Please add the type of negotiation strategy in the README file. All the tasks seem to be the same there.
- Make sure that the class distribution is not skewed towards some label.

kushalchawla:
Thanks for your reply. I will get to it asap.

Some comments:

- The context is missing since they are the first utterances of the conversation. I will describe what the context means in the definition. Thanks.

- other-need: when a participant uses other-need, they discuss a need for somebody else so that they themselves can get a better deal. This makes it a selfish strategy. I will clarify in the definition as well.

- Skewness: Oh is that a requirement? As you might expect, most utterances are non-strategic so the data currently is highly skewed towards the negative class. Ideally, I would prefer to reflect that in the data. But do you want me to sub-sample the instances from the negative class randomly so that both classes have the same number of instances? 

I will fix everything else, as per your comment.

yeganehkordi:
Thanks for your clarification.
I think skewness is a requirement. @danyaljj @swarooprm What do you think about this?

danyaljj:
Ya, especially for the classification tasks, we would prefer to have near-uniform skew.

kushalchawla:
Great. Thanks for clarifying that. I will subsample the negative class to keep it near uniform and also make other changes. Should be able to get back in a day or two. Thanks.

kushalchawla:
Hi @yeganehkordi @swarooprm . 

I was able to incorporate the suggestions. I request you to review asap. Thanks!

swarooprm (approved):
LGTM!
Congrats on your first task! Looking forward to your future PRs :)

PR #194, opened 2021-09-06
Task 360-362 : "Yes, and" response generation and classification.

Maitreyapatel:
These tasks are from [Grounding Conversations with Improvised Dialogues](https://arxiv.org/pdf/2004.09544.pdf) .
Please review it.


swarooprm:
@kuntalkumarpal, can you review these tasks?

tasks/README.md
```diff
@@ -238,4 +238,7 @@ Name | Summary | Category
 `task301_record_question_generation` | Given a passage, generate a fill-in-the-gap question based on it | Question Generation
 `task302_record_classification` | Given a passage and a question, classify the answer to the question based on the options | Classification
 `task303_record_incorrect_answer_generation` | Given a passage and a question, write an incorrect answer for the question | Incorrect Answer Generation
-`task339_record_answer_generation` | Given a passage and a question, answer the question based on the passage | Answer Generation
\ No newline at end of file
+`task339_record_answer_generation` | Given a passage and a question, answer the question based on the passage | Answer Generation
+`task360_SPOLIN_YesAnd_response_generation` | Given a prompt, generate the "yes, ands" response | Response Generation
+`task361_SPOLIN_YesAnd_prompt_response_classification` | Given a prompt and a response, classify whether the response is "yes, ands" type | Classification
```

Maitreyapatel:
Resolved the capitalization issue for task file names.

tasks/README.md
Maitreyapatel: Renamed the files.

swarooprm:
@Maitreyapatel 
Let's hear from @kuntalkumarpal regarding the reviews and then you can fix.

kuntalkumarpal:
@Maitreyapatel @swarooprm 
Task 362 : Definition
Given the prompt and **response pair**, classify the them to \"yes\" if response is \"yes, and\" type. Otherwise classify it as \"no\".

The Task needs only one response.

kuntalkumarpal:
Rest of the tasks look good to me

Maitreyapatel:
Hi @kuntalkumarpal ,
I think, there is misunderstanding here. In case of task362, there are two responses as input (i.e., "Response 1" and "Response 2") and model has to detect which one is appropriate.  Also, I think, you are referring to task361. I have modified the definition slightly avoid any ambiguity.

Thanks!

kuntalkumarpal:
@Maitreyapatel @swarooprm Looks good to me

swarooprm:
@Maitreyapatel, these are good tasks. Just make sure that labels are balanced. I saw in task 361 that, most of the answers are yes.

I was wondering if all the next dialogues in a conversation fall into this category of "yes and .." because, in instances of 360, I saw this pattern. The output does not always agree with the input but continues the conversation on that. Let me know what do you think about it?

Maitreyapatel:
@swarooprm I rechecked and for task361 labels are well balanced. Here, "yes" class has ~3100 and "no" class has ~3200 instances. 

Yes, they do fall in the category of "yes and ...". For example, agreeing with previous dialogue is definitely "yes, and ...", but without stating whether you agree with the previous dialogue also falls into this category as you are not opposing the fact. But if you oppose it by directly saying no or arguing the facts then this falls into opposite category .

swarooprm (approved):
LGTM

PR #197, opened 2021-09-07
Tasks 365-374 Synthetic Data and Tasks 517-520

kurbster:


danyaljj:
@nrjvarshney wondering if you can help with this PR's review?

swarooprm:
> @nrjvarshney wondering if you can help with this PR's review?

@nrjvarshney, pinging to check if you have seen my earlier message?

nrjvarshney:
* Task 365 - It is similar to task 157. In task 157, the input is a well-formed sentence but here it is a random string. 
* Task 366 - I'm not sure if the string " So this is a bad example" is required in the explanation. In the definition, you could specify how an empty list looks like ("[]"). The output here is str 
```
"output": [
                "[457, 941, 233, 137, 239]"
            ]
```
The order of instances should not make a difference here but if we are giving it as a string then the order also becomes important. 

* Task 367 - Same comment as Task 366. Also, you use "Seven and eleven" in the explanation and "7, 11" in the input. I think we should keep it consistent. 

* Task 368 - "Seven and eleven" comment from Task 367. It says "The output should be a list of integers that is the result of applying that logic to the input list." but the output is floating-point values also such as "0.5". Also comparing the floating-point values as strings could create problems because of rounding and number of decimal points used. Sometimes you use ".75" and "0.5" the other times. This is also a problem.

* Task 369 - "In this task you ...": comma after the word "task" is missing, Add the word "then" in the sentence "If every integer in the input list is odd an empty list should be returned.": "If every integer in the input list is odd then an empty list should be returned.". + same comments as above tasks. 

Will review the other tasks later.
cc @kurbster

kurbster:
I have changed task 365 to remove any vowels, so it differs from task 157.
For task 366, I don't understand what you mean by saying the output is a string. It is required to be a string. I don't know what you want me to do. Should I change it like this?
`"[457, 941, 233, 137, 239]" => "457, 941, 233, 137, 239"`. Every task I have done involving a list has been like this.
Task 367 I changed to use the numeric values not spelled out.
Task 368 I changed to say `The output should be a list of numbers`. I changed all decimals to include a leading 0. Also, when dividing I am only dividing by 4, thus the output will always be exact, 0.25, 0.5, 0.75 so rounding won't be an issue. I've added a line to explain not to round the output just to clarify.

I've applied all similar comments to other tasks. Like _So this is a good/bad example_ defining what an empty list is, and using numeric values instead of spelling numbers out.

kurbster:
Also, I was working on other tasks, Task 517-520. I forgot to create another branch and accidentally pushed those changes to this branch. @nrjvarshney could you review those as well? @danyaljj  Should I remove those other tasks and create a new PR or is it ok to combine them?

swarooprm:
Reviews of tasks 517-520: https://github.com/swarooprm/CSE576HW/pull/13

danyaljj:
I went through the new tasks and they seem good. 
Thank you, all!

PR #199, opened 2021-09-07
Task 376-378 Word length tasks

nrjvarshney:
`task376_reverse_order_of_words` |  Reverse the order of words in the given sentence | Answer Generation
`task377_remove_words_of_given_length` | Remove all words of a given length in the sentence | Answer Generation
`task378_reverse_words_of_given_length` | Reverse all words of a given length in the sentence | Answer Generation


danyaljj:
@kurbster windering if you can review this task?

kurbster:
> @kurbster windering if you can review this task?

Yes! I can start right now.

kurbster:
For Task 376, when I originally read the definition I thought the order of characters would be reversed (like task 378), maybe the definition could be more clear by saying "reverse the order that words appeared in the given sentence." You should add a negative example for this case, where each word is reversed, i.e "the quick brown fox" => "eht kciuq nworb xof". Also, in the current negative example the word _parking_ needs a quote after it to remain consistent. 

For Task 377, I feel the definition could be worded better, "Count of letters determine ..." => "The number of letters in a word determine its length." For your negative example I think you should add more explanation "The words 'in' and 'of' are length 2 however, the word 'of' is not removed. The answer should have been ..." Another negative example should be added where instead of removing all words of length 2, it removed 2 characters from all words, i.e "the quick brown fox" => "e ick own x"

For Task 378, same comment about the definition as Task 377. For your negative example you should explain why this is a negative example, "The word 'are' is reversed even though it is of length '3'. The answer should have been ..." Another negative example should be added where instead of reversing all words of length 2, it reversed 2 characters from all words, i.e "the quick brown fox" => "hte uqick rbown ofx"

kurbster:
Also, @danyaljj is it desirable to repeat the task in every instance? Every instance has "Reverse the order of words in a given sentence." at the end of the input. (Or whatever the prompt is for the respective task).

danyaljj:
> Also, @danyaljj is it desirable to repeat the task in every instance? Every instance has "Reverse the order of words in a given sentence." at the end of the input. (Or whatever the prompt is for the respective task).

Unless necessary, we prefer to keep all the task definitions/prompts in the "Definition" field.

kurbster:
For Task 376, the explanation for the second negative example include something like `The characters of each word are reversed, instead of the order of words being reversed.` 

For Task 377 per danyljj, you should remove the definition `Remove all words of length ...` from the input. For the second negative example, the explanation should include with something like `Two characters were removed from every word, instead of removing words of length 2.`

For Task 378, you should also remove the definition `Reverse all words of length...` from the input. For the second negative example, the explanation should include with something like 'However, every word was reversed` 

Basically we want to explain why those negative examples are negative. You can say this however you feel is best, you just need to provide context as to what specifically made the negative examples bad.

nrjvarshney:
In Task 377 & 388 , there is a parameter involved. So, eliminating the sentence " Remove all words of length \'{len}\' in the given sentence." from the input would make the input incomplete.

kurbster:
@nrjvarshney that makes sense. I'm wondering if we could phrase it better.  @danyaljj what do you prefer for the input to the instances? 

1. `Length: '2'. Sentence: 'the quick brown fox jumped over the lazy dog'.`
2. `Sentence: 'the quick brown fox jumped over the lazy dog'. Remove all words of length '2' in the given sentence.`

danyaljj:
Either way is fine. As long as it is clear to an average human reader, we're good!

kurbster:
Ok, then everything is good. Ready to merge!

kurbster (approved):

danyaljj:
Thanks to both of you!

PR #202, opened 2021-09-07
Task 383 and Task 456-459: MATRES

yeganehkordi:
Based on: http://cogcomp.org/papers/NingWuRo18.pdf


tasks/task383_matres_classification.json
```diff
@@ -0,0 +1,11564 @@
...
+    "Contributors": [
+        "Yeganeh Kordi"
+    ],
+    "Source": [
+        "MATRES (http://cogcomp.org/papers/NingWuRo18.pdf), (https://github.com/CogComp/MATRES)"
+    ],
+    "Categories": [
+        "Classification" 
+    ],
+    "Definition": "We want to determine which verb can be anchored in the time axis, i.e., has a well-defined notion of time. You will be given a question and a context separated with a \n, and you have to answer that the given verb can be anchored in time or not. We say a verb can be anchored in the real timeline if and only if an Event happened in the past or is happening now or is unconditionally planned to happen in the future. If the verb is not anchorable, choose one of the No. It's intention/wish/opinion, No. It's hypothetical/condition, No. It's negation, No. It's abstract/non-specific, No. It's static or No. It's recurrent. And then classify your answers into \"yes_its_anchorable\", \"no_its_intentionwishopinion\", \"no_its_hypotheticalcondition\", \"no_its_negation\", \"no_its_abstractnonspecific\", \"no_its_static\" and \"no_its_recurrent\".",
```
danyaljj: - I think you can safely drop "We want to determine which verb can be anchored in the time axis, i.e., has a well-defined notion of time." 
- "a verb can be anchored  if and only if an Event ....": this sentence fails to say how the discussed verb and event are connected to each other. 
- About output labels: the description clearly defines what is anchorable and what is not. But I am not sure if the fine-grained labels (intention/wish/opinion, hypothetical/condition, negation, etc.) are clear.

yeganehkordi: Thanks! I updated the definition.

tasks/task383_matres_classification.json
```diff
@@ -0,0 +1,11564 @@
...
+    "Positive Examples": [
+        {
+            "input": "It was followed by Denmark , Norway , Germany , France , Greece , Luxembourg , Spain , Britain , the United States , Iceland , Belgium , Italy , Portugal and Turkey . NATO decided at last year 's Madrid summit to (invite) the three eastern European countries to start accession talks .  \n Can the verb (invite) be anchored in time?",
+            "output": "no_its_intentionwishopinion",
+            "explanation": "This is a good example. The \"invite\" is not anchorable and It's intention."
+        },
+        {
+            "input": "In Washington today, the Federal Aviation Administration (released)air traffic control tapes from the night the TWA Flight eight hundred went down.  \n Can the verb (released) be anchored in time?",
+            "output": "yes_its_anchorable",
```
danyaljj: why "yes_its_anchorable" and not "Yes, it is anchorable"?

tasks/task383_matres_classification.json
```diff
@@ -0,0 +1,11564 @@
...
+        "Classification" 
+    ],
+    "Definition": "You will be given a question and a context separated with a \n, and you have to answer that the given verb can be anchored in time or not. We say a verb can be anchored in the real timeline if and only if an verb happened in the past or is happening now or is unconditionally planned to happen in the future. If the verb is not anchorable, provide more detail by choosing one of the other labels. If verb is about a intention, wish or opinion in the future but it's not unconditionaluse No. It's intention/wish/opinion, if verb implies a condition or hypothetical choose No. It's hypothetical/condition this verbs usually come after if, choose No. It's negation if verb is about an event that is not going to happen, if the verb doesn't limited to a specific time choose No. It's abstract/non-specific, If a verb always happens use No. It's static and if it's about a repetitive event use No. It's recurrent. And then classify your answers into \"yes_its_anchorable\", \"no_its_intentionwishopinion\", \"no_its_hypotheticalcondition\", \"no_its_negation\", \"no_its_abstractnonspecific\", \"no_its_static\" and \"no_its_recurrent\".",
```
danyaljj: typo: "unconditionaluse"

danyaljj: Would it be easier to read if you revise 
> .... No. It's negation. 

 to 
> .... "No, It's negation" ....

?

tasks/task383_matres_classification.json
yeganehkordi: I used the exact labels of the dataset. I can change them now.

tasks/task383_matres_classification.json
yeganehkordi: Thanks! I'll fix them now.

yeganehkordi:
@danyaljj Thanks for your comments! I changed the labels and updated the definition.

tasks/task383_matres_classification.json
```diff
@@ -0,0 +1,11558 @@
...
+    ],
+    "Definition": "You will be given a question and a context separated with a \n, and you have to answer that the given verb can be anchored in time or not. We say a verb can be anchored in the real timeline if and only if an verb happened in the past or is happening now or is unconditionally planned to happen in the future. If the verb is not anchorable, provide more detail by choosing one of the other labels. If verb is about a intention, wish or opinion in the future but it's not unconditional use No, It's intention/wish/opinion, if verb implies a condition or hypothetical choose No, It's hypothetical/condition this verbs usually come after if, choose No, It's negation if verb is about an event that is not going to happen, if the verb doesn't limited to a specific time choose No, It's abstract/non-specific, If a verb always happens use No, It's static and if it's about a repetitive event use No, It's recurrent. And then classify your answers into \"Yes, it is anchorable\", \"No, it is intention/wish/opinion\", \"No, it is hypothetical/condition\", \"No, it is negation\", \"No, it is abstract/nonspecific\", \"No, it is static\" and \"No, it is recurrent\".",
```
danyaljj: I am still a bit fuzzy on the definition. Am I correct that this is a whole sentence? In which case, it is a very complex and long sentence (besides its grammatical issues). In that case, it might be helpful to try to break it into several simpler sentences. 

> If verb is about a intention, wish or opinion in the future but it's not unconditional use No, It's intention/wish/opinion, if verb implies a condition or hypothetical choose No, It's hypothetical/condition this verbs usually come after if, choose No, It's negation if verb is about an event that is not going to happen, if the verb doesn't limited to a specific time choose No, It's abstract/non-specific, If a verb always happens use No, It's static and if it's about a repetitive event use No, It's recurrent.

yeganehkordi: Sorry, it wasn't supposed to be one sentence. I updated the definition and hope it's better now.

danyaljj:
Thanks! I think that we still need to improve the task definition to make it more concise and crisp. 
It might be helpful to second on this PR. @kurbster wondering if you can review this PR, whenever you get a chance.

yeganehkordi:
What if we focus on that a verb is anchorable or not and just use "yes" and "no" labels? I think this makes the task more concise.
Actually, we can break it into many sub-tasks and have a yes-no classification for each category.

danyaljj:
Ya, I think that can help!

kurbster:
> Thanks! I think that we still need to improve the task definition to make it more concise and crisp.
> It might be helpful to second on this PR. @kurbster wondering if you can review this PR, whenever you get a chance.

I will take a look.

kurbster:
I don't fully understand what `unconditionally planned to happen in the future` means. Does it mean it is always guaranteed to happen, or someone is expecting it to happen? Which one is correct?

1. I am planning a surprise birthday party for my friend. Verb is `planning`
2. I am going to plan a surprise birthday party for my friend. Verb is `plan`
3. He is building a new house next month. Verb is `building`
4. Eventually the universe will explode in a heat death. Verb is `explode`

In the first, and second example I have unconditionally planned for the party to happen. However, if I procrastinate later and forget to plan, then the party won't happen. So these verbs are conditional on the fact I will actually follow through. For the third example, he will be building a new house next month, is stated unconditionally that this will happen. However, if he dies, he will not build the house.

Whereas the fourth example will always happen. We should specify what unconditional planning means. Does it mean in the current moment this is planned to happen. Or this will always happen no matter what?

Also, maybe change the last line `The output should be "Yes" if the verb can be anchored in time and "No otherwise`. This makes more sense to me personally, but maybe is repetitive and unnecessary with respect to GPT3.

Some grammatical fixes. Change an into a. Add commas to the list.

We say a verb can be anchored in the real timeline if and only if **_a_** verb happened in the **_past,_** is happening **_now,_** or is unconditionally planned to happen in the future.

yeganehkordi:
`unconditionally planned to happen in the future` means it is guaranteed to happen in the future. If someone is expecting it to happen, it's a wish, opinion, or intention, and it's not anchorable anymore. 
In your examples, 1 is anchorable because you are planning now and the fact that party may not happen doesn't matter, but 2, and 3 are not anchorable; they are intended. 
For example, in `I (want) to (leave) here` want is anchorable but leave is an intention. Because the event of wanting won't change, but I may not leave here for some reasons.
Thanks, I will update the definition now.

yeganehkordi:
@danyaljj @kurbster I've added four new sub-tasks from this dataset. Could you please review them?

kurbster:
Excellent job with separating the tasks! Much easier to comprehend now.

For task383, change the definition to start like this `You will be given a context and a verb ...`. For the second positive example and first negative example, you should add _the_ so `In this sentence, the verbing action happened ... `

For task456, add a space in the first negative example explanation here `intention, wish, or opinion`. 

For task 458, in the first task you put parentheses around the verb `(claim)`. You used quotes around the verbs for every other task explanation, for consistency you should use quotes here.
  
For consistency across the tasks:
1. In task 45[6-9] the definition says `You will be given a verb and a context ... ` but, the context is passed first. I doubt this matters but from a consistency point you should change it to the correct order. 
2. For task 383 you end each sentence with `Can the verb (verb) ...` whereas in every subsequent task you end the sentence with `Verb: `. Personally, I think the second style is more concise and inline with the intent of this project. Running this `sed` command will fix it.
`sed -e 's/Can the verb (/Verb: /' -e 's/) be anchored in time?//' task383_matres_classification.json`
3. In the explanations for all examples, if you say `In this sentence, verbing action happened ...` change it to `In this sentence, the verbing action happened ...`
4. Also, to be picky I think changing the definition like this would make more sense. The new tasks are asking _if_ the verb is a  _yes_ or _no_ answer
`you have to answer that the given verb ... `
`you have to answer if the given verb ... `

yeganehkordi:
Thanks for your comments! I updated the tasks.

danyaljj:
@kurbster give the changes a look?

kurbster:
I reviewed the changes everything looks good. Ready to merge!

kurbster (approved):

danyaljj:
Thank you, both!

PR #205, opened 2021-09-08
Update test_all.py: check the types of several recently added keys. 

danyaljj:
We must fix the root of the failure before merging this PR.


danyaljj:
@eshaanpathak wondering if you can help with the review of this PR?

PR #209, opened 2021-09-08
Task 391-393 from cod3s paper

Palipoor:
This was a human evaluation, with outputs in a range from 0 to 5 for different extents of plausibility. I took an average from human inputs, then converted it to a binary classification task.


danyaljj:
Can you expand on "took an average from human inputs"? (like, multiple human annotators labeled each instance?) 

@amirrezamirzaei wondering if you can review this PR?

amirrezamirzaei:
looks good! minor comment:
1- in task 391&392 definition can you explain more about the not_decidable group and its difference from the not_plausible group. also add a positive example for not_decidable.
2- can you add more instances of not_decidable to 392 so we have around the same amount of each answer.
3- also can you use the same order we use for every task json fields (Contributors-Source-...).

tasks/README.md
```diff
@@ -296,3 +296,6 @@ Name | Summary | Category
 `task380_boolq_yes_no_question` | Given a passage and a yes/no question, answer the question based on the passage | Answer Generation
 `task381_boolq_question_generation` | Given a passage, generate a yes/no question that can be answered based on the passage | Question Generation
 `task382_hybridqa_answer_generation` | Given a question about part-of-speech tag of a word in the question, answer the question | Answer Generation
+`task391_causal_relationship` | Given two sentences, decide whether there is a cause and effect relationship between them, and the second one is a result of the first one. | Answer Generation
```
amirrezamirzaei: this could be written as
`Given two sentences, decide whether the second sentence can be the result of the first one`

tasks/README.md
```diff
@@ -296,3 +296,6 @@ Name | Summary | Category
...
+`task392_inverse_causal_relationship` | Given two sentences, decide whether there is a cause and effect relationship between them, and the first one is a result of the second one. | Answer Generation
```
amirrezamirzaei: also here

Palipoor:
> Can you expand on "took an average from human inputs"? (like, multiple human annotators labeled each instance?)
> 
> @amirrezamirzaei wondering if you can review this PR?

Yes. Multiple humans had annotated each pair, and I took an average of them.

Palipoor:
> looks good! minor comment:
> 1- in task 391&392 definition can you explain more about the not_decidable group and its difference from the not_plausible group. also add a positive example for not_decidable.
> 2- can you add more instances of not_decidable to 392 so we have around the same amount of each answer.
> 3- also can you use the same order we use for every task json fields (Contributors-Source-...).

I removed the not-decidable category, it was vague. I also fixed other issues. thanks!

amirrezamirzaei:
thanks look good to me one more minor comment!
why did you use both Answer generation and classification as categories for 391&392?  I think both of them fit better in the classification category but I'm not sure.

Palipoor:
> thanks look good to me one more minor comment!
> why did you use both Answer generation and classification as categories for 391&392? I think both of them fit better in the classification category but I'm not sure.

I don't know, because I had a classification task, but other contributors thought it'd match "answer generation" better. That's why I put both.

amirrezamirzaei:
I think it's better if you use just classification.

amirrezamirzaei:
LGTM , just fix the task type in the ReadMe file.

danyaljj:
Thank you both! Merging!

PR #210, opened 2021-09-08
Task 394-396: PersianQA

yeganehkordi:
Based on the: https://github.com/sajjjadayobi/PersianQA


danyaljj:
Since this is a non-English task, please add the appropriate fields based on #113. 

@Palipoor wondering if you can review this PR?

yeganehkordi:
In most of the cases, the output language is Persian, but there are some questions asking about a scientific term, etc, and their answer is in English. Is it okay to have two output languages?

danyaljj:
I think we should go with the single dominant language, which in this case is "fa".

yeganehkordi:
Thanks!

Palipoor:
The tasks look good to me! 
I'm not sure about the descriptions though. Should they be in English? (Just asking to know for future tasks)

yeganehkordi:
Thanks! Yes, descriptions and explanations should be in English.

danyaljj:
Thanks! Merging!

PR #212, opened 2021-09-08
task 397-399 : Semeval2018 task1 

amirrezamirzaei:
task 397-399 using [semeval 2018 task 1](https://aclanthology.org/S18-1001.pdf)
(I will create more task from this dataset)


swarooprm:
@yeganehkordi, wondering if you can review these tasks?

yeganehkordi:
Sure! I'll review it now.

tasks/task397_semeval_2018_task1_tweet_anger_detection.json
```diff
@@ -0,0 +1,15662 @@
...
+            "input": "Caleb had a nightmare about zombies. I had a dream about freedom.......",
+            "output": "NOT",
+            "explanation": "The author of the tweet isn't angry."
+        }
+    ],
+    "Negative Examples": [
+        {
+            "input": "This time last year I was pissed at the top of the Eiffel Tower. Today I was down a manhole in the rain. Happy Anniversary @Jordannea11 \ud83d\ude18\u2665\ufe0f",
+            "output": "ANGRY",
+            "explanation": "the tweet author isn't angry right now so the output should be \"not\"."
```
yeganehkordi: - Is there any reason that you used "ANGRY" instead of "Angry"? 
- I think using "Not angry" label is better.
- "... so the output should be \"not\". ": Labels in the explanation should be same as definition. (i.e., "NOT" or "Not angry")
- I suggest using json.dumps(obj, ensure_ascii=False) to have emojis in JSON. Their encoded version isn't human readable.

tasks/task398_semeval_2018_task1_tweet_joy_detection.json
```diff
@@ -0,0 +1,10808 @@
...
+        {
+            "input": "'What has you smiling like that'\\n'Dominos sent me a coupon' \ud83d\ude2c",
+            "output": "HAPPY",
+            "explanation": "The tweet author is happy because she has received a coupon."
+        }
+    ],
+    "Negative Examples": [
+        {
+            "input": "Time for bed in a real bed after four nights camping. #bliss",
+            "output": "ANGRY",
```
yeganehkordi: In addition to comments of previous task, if using "ANGRY" label is intentional here, it would be helpful to mention that we are not allowed to use other labels.

amirrezamirzaei: thanks, it wasn't intentional.

tasks/task397_semeval_2018_task1_tweet_anger_detection.json
```diff
@@ -0,0 +1,15662 @@
...
+        {
+            "input": "sick of this shit. #mad #angry. Rowan Atkinson Is Not Dead. Just A Bloody Online Hoax😡😡😡😡🤬🤬🤬🤬",
+            "output": [
+                "Angry"
+            ]
+        },
+        {
+            "input": "@lucybeecoconut A make up remover and insect sting relief!",
+            "output": [
+                "Not"
```
yeganehkordi: Could you use the same label in the definition and instances?

yeganehkordi (commented):
Looks good to me!
Make sure that the class distribution is not skewed towards some label. 
Also, I think it's more ideal to have one PR per each dataset.

amirrezamirzaei:
I think the class distribution is balanced.
task 397 contains 1509 angry instances, 1095 not angry instances.
task 398 contains 999 happy instances, 796 not happy instances.
task 399 contains 1741 sad instances, 1161 not sad instances.

swarooprm (commented):
Looks good to me. 
One final suggestion: In task 397, negative examples have label "not angry" whereas in the case of instances, the label becomes "not". Could you fix and make labels consistent?

swarooprm (approved):
LGTM

PR #213, opened 2021-09-09
Update test_all.py

danyaljj:
Adding a constraint for making sure that all the instances have at least one output label.


swarooprm (approved):
LGTM

PR #215, opened 2021-09-09
task 400 : PAWS

amirrezamirzaei:
task 400 using [PAWS](https://arxiv.org/abs/1904.01130)


danyaljj:
Looks good to me!  Leaving it here for @swarooprm to check before merging it.

swarooprm (approved):
Looks good to me! 
@amirrezamirzaei, I think you can create a similar task of "text modification" category where input is a given sentence and output is multiple paraphrases of it.

swarooprm:
Please rename your file with lowercase letters and also in readme so that tests pass and we can merge this PR @amirrezamirzaei

amirrezamirzaei:
@swarooprm, I can add the task you suggested but the paw dataset doesn't contain that many multiple paraphrase for each instance.
here how many paraphrases it has:
8 paraphrase: 15 instance 
7 paraphrase: 10 instance 
6 paraphrase: 11 instance 
4 paraphrase: 912 instance 
3 paraphrase: 1171 instance 
2 paraphrase: 7471 instance 
1 paraphrase: 35449 instance 
should I go ahead and just make the instances from those with more than 3 paraphrase? I kinda feel the output count would be too low.

PR #216, opened 2021-09-09
Fix test_all

Mehrad0711:


danyaljj:
I think this is a good change. The issue is that some tasks don't have enough examples, which I think they should. 
For example, currently `task002_quoref_answer_generation` is failing the tests. @swarooprm what do you think we add a few examples?

swarooprm:
> I think this is a good change. The issue is that some tasks don't have enough examples, which I think they should.
> For example, currently `task002_quoref_answer_generation` is failing the tests. @swarooprm what do you think we add a few examples?

Yes, good initiative, however currently we have lots of tasks where this test fails and so this needs major revision. I think we can take this up later when we revise examples to add diverse positive and negative examples (currently no attention has been given to the diversity of examples).

PR #224, opened 2021-09-10
Enhancements to README and test_all.py

eshaanpathak:
1. Added a check to `src/test_all.py` to ensure that the `.json` file extension ending is not present in `tasks/README.md`. I then went through and removed these extensions so they would pass this test.
2. Fixed the tasks to be in numerical order in `tasks/README.md`.
3. Edited grammar in the description of each task in `tasks/README.md`.
4. Added a check to make sure all characters are lowercased in the filename for consistency throughout filenames. I then went through and lowercased all task filenames so they would pass this test. 
5. There appears to be different tasks under the same task numbers for tasks 119 and 184. You can fix this by renaming them to the last 2 tasks after task submissions are no longer accepted.
6. There are also a lot of gaps between task numbers. Are you going to go by hand and renumber everything after task submissions are no longer accepted?

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -103,7 +103,7 @@ Name | Summary | Category
 `task102_commongen_sentence_generation` | Given a collection of concepts, use them in a coherent sentence. | Sentence Generation
 `task103_facts2story_long_text_generation` | Given 5 facts, write a story that incorporates them. | Long Text Generation
 `task104_semeval_2019_task10_closed_vocabulary_mathematical_answer_generation`	|  Answering multiple choices mathematical problem described with a closed-vocabulary.	| Answer Generation, Arithmetic
-`task105_story_cloze-rocstories_sentence_generation` | Given a four sentences, predict the next (fifth) coherent sentence. | Sentence Generation
+`task105_story_cloze-rocstories_sentence_generation` | Given four sentences, predict the next coherent sentence. | Sentence Generation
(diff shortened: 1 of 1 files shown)


src/test_all.py
```diff
@@ -64,19 +64,26 @@
...
+            true_file = file.replace(".json", "")
+            for char in true_file:
+                if char.isupper():
+                    raise Exception(f" * Looks like there is an uppercase letter in `{true_file}`. All letters should be lowercased.")
+
+            if file in task_readme_content:
+                raise Exception(f" * Looks like the .json file extension ending is present with the task name in `tasks/README.md` when it should just be `{true_file}`")
+
+            if true_file not in task_readme_content:
+                raise Exception(f' * Looks like the task name `{true_file}` is not included '
```
danyaljj, line 86: Cool! 👏 👏

danyaljj:
Thanks, @eshaanpathak ! This is super helpful! 

> There appears to be different tasks under the same task numbers for tasks 119 and 184. You can fix this by renaming them to the last 2 tasks after task submissions are no longer accepted.
> There are also a lot of gaps between task numbers. Are you going to go by hand and renumber everything after task submissions are no longer accepted? 

Yeah, I think we should sort them out later when things cool down.

danyaljj (approved):

danyaljj:
Leaving it here for @swarooprm to check before we merge it.

swarooprm (approved):
LGTM! Thanks @eshaanpathak. This is super-helpful.

PR #226, opened 2021-09-11
Task 405 from NarrativeQA

Palipoor:
Will add more tasks from this dataset.


danyaljj:
@shailaja183 wondering if you can review this PR?

shailaja183:
Task 405 looks good mostly, just a couple of minor comments;
1. Negative example explanation - instead of "two-part question", I would suggest to write "do not combine multiple questions"
2. Negative example explanation says that "The names in parantheses are the names of the actors and they shouldn't be in your question. Only ask questions with character names." -- also add this in the definition 
3. Typos - charracters, parantheses

danyaljj:
@shailaja183 check if the changes look good to you?

shailaja183:
looks good to me!

swarooprm:
@Palipoor: this task looks good and I am ready to merge as soon as the following comment gets addressed:
In the definition, replace "questions" with "a question". We expect model to generate a single question, so let's say that.

swarooprm:
Suggested Definition: 
"You will be given a summary of a story. You need to create a question that can be answered from the story. You can create a question about characters, events, facts and beliefs, etc. Your question should be specific, try not to use pronouns instead of full names. As the stories are sometimes movie plots, they will contain actor names in parentheses. You should not use those names. Only use character names. Try to ask a question about all parts of the plot, not just the beginning."

I tried changing it in your branch, but I don't have push access.

swarooprm:
Also, you can create other tasks from NarrativeQA such as answer generation.

Palipoor:
@swarooprm I changed the definition. Sure, I will create other tasks as well!

swarooprm:
> @swarooprm I changed the definition. Sure, I will create other tasks as well!

Great! That was quick!

swarooprm (approved):

PR #227, opened 2021-09-11
Tasks 406-417 : Mickey tasks

Maitreyapatel:
Please review these language perturbations tasks.

### DIFF
--- a/tasks/task408_mickey_it_sentence_perturbation_generation.json
+++ b/tasks/task408_mickey_it_sentence_perturbation_generation.json
@@ -18,7 +18,7 @@
         {
             "input": "Puoi usare una barca per attraversare l'acqua.",
             "output": "È possibile utilizzare una barca per attirare l'acqua.",
-            "explanation": "The output sentence is semantically related to the input as both are taking about the boat sentence and follows the commonsense. Hence, the answer is correct."
+            "explanation": "The output sentence is semantically related to the input sentence as both are talking about the boat and follows the commonsense. Hence, the answer is correct."
         }
     ],
     "Negative Examples": [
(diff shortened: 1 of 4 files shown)


eshaanpathak:
In the explanation section of the positive and negative examples, you should also include a translation in English using Google Translate to see what is being said, so it's easier to understand how the perturbations are semantically and syntactically similar or not.

Otherwise, to me it looks like there's a large amount of perturbations that aren't covered by the output section in the instances, so it might be hard to make the model get one or more of the "correct" outputs. Hopefully this isn't the case after you add the translations.

Maitreyapatel:
Thanks for reviewing these tasks. 
I believe that model should learn these settings without any explicit translation provided. But surely I can add some differences in explanation in English to guide the model better.

eshaanpathak:
You definitely should for the positive and negative examples.

swarooprm:
Overall, I like these tasks. However, I did not understand "And the generated sentence should have high commonsense plausibility.", may be you can elabore a bit more in the definition. 
Also, my understanding from the explanation is that: the only difference is negative examples do not involve commonsense. Maybe, you can elaborate a bit more in the negative example explanation by providing a suggestion on what could have been a good examplation
Minor: Category should be text modification here.

Maitreyapatel:
Thanks @swarooprm for suggestions. I will incorporate these suggestions as soon as I can.

swarooprm:
> Thanks @swarooprm for suggestions. I will incorporate these suggestions as soon as I can.

@Maitreyapatel, wondering if you have any update here?

Maitreyapatel:
@swarooprm Please refer the updated changes in all positive/negative example explanations and updated task definition.
Please let me know if you have any further suggestions to improve them.

swarooprm:
Looks good to me. I am ready to merge, but I saw this small typo: "as both are talking about what to become if one doesn’t to become jury"-->" as both are talking about what to become if one doesn’t want to become jury"

I would suggest to you have a grammar check either manually or via grammaraly, and this PR is good to get accepted. @Maitreyapatel

Maitreyapatel:
Thanks for noticing this issue. I have checked the grammar and modified the minor issues.

swarooprm (approved):
LGTM

PR #228, opened 2021-09-11
Fix type keys

amirrezamirzaei:
fixing the type of some keys(source, contributors, categories) so #205 can be merged.


danyaljj:
Thank you @amirrezamirzaei ! This is very helpful!

PR #231, opened 2021-09-11
Task 424-427: HindiEnglish Corpora

Mirzyaaliii:
task424_HindiEnglish_Corpora_hi_en_translation -> Given a Hindi sentence, convert it into English.
task425_HindiEnglish_Corpora_en_hi_translation -> Given an English sentence, convert it into Hindi.
task426_HindiEnglish_Corpora_hi-en_classification -> Given a Hindi sentence and its corresponding English sentence, classify whether it is correct or not.
task427_HindiEnglish_Corpora_hi-en_language_identification -> Given a sentence, identify whether it is in Hindi or English.

Please review.


eshaanpathak:
You should explain why the positive examples are correct instead of just stating they are.

How does this differ from #233? These PRs both look the same to me.

Mirzyaaliii:
> You should explain why the positive examples are correct instead of just stating they are.
> 
> How does this differ from #233? These PRs both look the same to me.

Thanks for the comments.

- Changes are done as suggested.

- Tasks of #233 and these tasks are the same, just the dataset is different. Here, [HindiEnglish Corpora](https://www.kaggle.com/aiswaryaramachandran/hindienglish-corpora) is used and in #233 [Asian Language Treebank (ALT)](https://www2.nict.go.jp/astrec-att/member/mutiyama/ALT/) is used.
And same tasks are allowed with different datasets.

eshaanpathak:
I see. That makes sense. Here are some more reviews:

1. There's a typo with "sentencce" in task 424.
2. For the explanations, it might be best to just paraphrase the English output into your own words. For the second positive example of task 424, something like "This Hindi sentence is properly translated into English because the Hindi sentence also talks about how a very small amount of hydrogen sulphide in water can have the smell of a rotten egg." would work as an explanation in my opinion.
2. There's a typo with "unrelible" in the first negative example of task 424.

Mirzyaaliii:
> I see. That makes sense. Here are some more reviews:
> 
> 1. There's a typo with "sentencce" in task 424.
> 2. For the explanations, it might be best to just paraphrase the English output into your own words. For the second positive example of task 424, something like "This Hindi sentence is properly translated into English because the Hindi sentence also talks about how a very small amount of hydrogen sulphide in water can have the smell of a rotten egg." would work as an explanation in my opinion.
> 3. There's a typo with "unrelible" in the first negative example of task 424.

Thank you very much for the comments!!

All the changes are done.

eshaanpathak:
One last thing: edit "is a sentence case" to "is in sentence case" in the explanations of the positive examples for `task424_hindienglish_corpora_hi_en_translation.json`.

@danyaljj @swarooprm After this change is made, these will look good to me if you want to take a quick look before you merge.

Mirzyaaliii:
> One last thing: edit "is a sentence case" to "is in sentence case" in the explanations of the positive examples for `task424_hindienglish_corpora_hi_en_translation.json`.
> 
> @danyaljj @swarooprm After this change is made, these will look good to me if you want to take a quick look before you merge.

Thanks for your valuable comments.

Changes are done.

swarooprm (approved):
LGTM.
@Mirzyaaliii, task426 is more like a transformation which we can apply to any tasks where just the input format changes, everything around that remains almost the same. Please create an issue "input perturbation" referencing this PR and hopefully someone can take up this to apply across tasks in our repo. This is a good simple transformation.

PR #232, opened 2021-09-11
Task 428-431 : Sent eval

amirrezamirzaei:
task 428-431 using [Sent Eval](https://github.com/facebookresearch/SentEval/tree/main/data/probing)


yeganehkordi (commented):
Thanks! 
- Task 428: Looks good to me!
- Task 429: I think ` the main verb of the sentence` might be unclear. Is it possible to explicitly mention the verb in the input? 
      In the first example, shone in `She shone her light around the space` seems to be in the past tense.

- Task 430: Looks good! It would be better if you can add another negative example.
- Task 431: Please replace the `objects of the sentence is` with the `object of the sentence is` or `objects of the sentence are`. 
     In some instances there are more then one object in the text but we have only one output. For example, in line 542:
     `She picked up her handbag, grabbed the box with the candle and oil, and headed to her car` 
    Can you mention the object which you are asking about, in the input?

amirrezamirzaei:
in regard to task 429: I agree with you but the raw data doesn't explicitly mention the verb. But I think it's clear which verb is the main verb in instances containing more than one verb.
in regard to task 431: I changed the description to `object of the main clause`.

yeganehkordi (commented):
Sounds good!
Please start explanations with a capital letter in all the tasks. And check tense of `shone` in the first example of task 429.

amirrezamirzaei:
ragarding task 429 : shone is labeled as past tense in the source dataset which is incorrect. should I look at all the other instances too?

yeganehkordi:
The past tense is correct. If you mean it's labeled as present tense; I think just correcting the examples is enough. Probably most of the datasets have some mistakes. If you have extra time, you take a glance.
@danyaljj @swarooprm Is it ok if there are some mistake labels in the original dataset?

yeganehkordi (approved):
Thanks! Looks good to me!

swarooprm:
@danyaljj, this is a relatively hard task for humans to do.
Many of the tasks revolve around the main clause, finding which is not trivial. We can merge these tasks, but going forward, I think we should have 2 broader categories: (i) hard for humans and (ii) easy for humans. 
@amirrezamirzaei, I can see that more tasks can be created which will be easy for humans (you can create an issue if you don't plan to create these tasks and someone else can pick it up):
(i) given a word, say if it singular or plural
(ii) given a word, say it's tense.
(iii) generating subject of a sentence
(iv) generating object of a sentence
(v) generating verb of a sentence.
(vi) creating more tasks like task 429 synthetically similar to https://github.com/allenai/natural-instructions-expansion/blob/master/tasks/task089_swap_words_verification.json

Regarding annotation issue: fix whatever you can see or find automatically. Ensure that most labels are correct. We can't do anything beyond this.
Overall, I think we should ACCEPT these tasks and I think we should have a new task category called 'Grammar' for these tasks.

amirrezamirzaei:
the raw data doesn't contain the actual word its classifying. for example in task 429 (detecting main verb tense) the instances are just classified as present and past, and the main verb isn't specifically mentioned.

danyaljj:
Yeah, the tasks are not super easy, but they're not too difficult either. I personally find it easier than some of our, say, summarization or title generation tasks. Overall, I am **for** these tasks. 

All the tasks definitions looked okay to me. Here is one suggestion: 

> In this task you are given a sentence. You must judge whether two consecutive tokens within the sentence have been inverted or not. Label the instances as \"Inversion\" or \"Original\" based on your judgment. 


 * "whether two consecutive tokens" is not clear refers to what two tokens. It might be better to say "whether **there exist** two consecutive tokens ..."
 *  "inverted" is an ambiguous term and can be confusing (note, "abc" can be an inverted form of "cba", which is not what you mean). It might be better to say "two consecutive words with flipped **orders/positions**", or sth along that line.

danyaljj:
Looks good, thank you all!

PR #241, opened 2021-09-12
Task 467-468: ParsiNLU-reading_comprehension

yeganehkordi:


gkaramanolakis:
Looks good to me! I would suggest changing "Given a passage, generate a question based on it" to "Given a passage, generate a question based on the passage" in the README file. Both JSON files look good to me.

yeganehkordi:
Thanks! I updated the README file.

gkaramanolakis (commented):
Thanks, LGTM!

PR #243, opened 2021-09-12
Task 473-474: ParsiNLU-multiple-choice

yeganehkordi:


Palipoor:
Hi. I think it's better if the task 473 is titled mc question answering, and not classification.

yeganehkordi:
Hi, Thanks for reviewing!
Actually, I'm not sure about the category. I used classification based on Swaroop's comment here:  #172 

> The answer generation task is good, but its mostly a classification task. You can have another task to generate answer (here answer is (A), (B) etc. ,so classification, there the answer will be pure text generation). For example, look at task004 or task002

swarooprm:
We actually had 'answer generation' category which was used when we used to generate text. But, clearly, we don't generate text here. MCQ is the most appropriate category here, feel free to create such a category if it does not exist.

@Palipoor, wondering if you can check and let @yeganehkordi know if you have any other feedback on this PR?

yeganehkordi:
Thanks! I changed the category to the "Multiple-Choice Question". I think we can split this category into Multiple-Choice Question generation and Multiple-Choice Question answering.

danyaljj:
Looks good, thank you all!

PR #244, opened 2021-09-13
Task 475: Yelp Polarity Classification

gkaramanolakis:
Classify a given Yelp review to positive or negative sentiment. 

The class distributions are: 
POSITIVE    3263
NEGATIVE    3237


yeganehkordi (commented):
Thanks! Looks good to me.

PR #262, opened 2021-09-14
Task 515-516 : Senteval

amirrezamirzaei:
task 515 and 516 using [Sent eval](https://github.com/facebookresearch/SentEval/tree/main/data/probing)


Palipoor:
Tasks look good to me!

danyaljj:
Some suggestions: 


 * At the end of " ... whether a single noun or verb has been replaced with another word with the same part of speech." we should elaborate what is the effect of such change (which a reader must use to solve the task): something, like, the resulting sentence sounds unnatural. 


*  "a sentence with one coordinating conjunction." include a list to help define it: "(for, and, nor, but, or, yet, so.)". 

 * "... whether the order of two coordinated clausal conjoints have been inverted or not.": like the first case, add a sentence the describes what signal one should look for to find the "inversions". I suppose an inverted combination should sorta sound out of place, compared to their un-inverted counterpart.

danyaljj:
Looks good, thanks!

PR #263, opened 2021-09-14
Task 521 - classification from QANTA

Palipoor:
I added this task from QANTA (https://people.cs.umass.edu/~miyyer/qblearn/index.html) 
I wanted to add the question answering task too, but it's a trivia and needs external knowledge. Is it ok to do this? @swarooprm @danyaljj


swarooprm:
I think we had some task from QANTA already in the repo. please check.
Also, seems like we have tasks from 517-520 already https://github.com/allenai/natural-instructions-expansion/pull/197

Palipoor:
I couldn't find the other task from QANTA. I renamed the file though.

swarooprm (commented):
This task looks good.
Some minor comments.
- The text is talking about anauthor and their works, so it belongs to theliterature category-->The text is talking about an author and their works, so it belongs to the literature category
- Also I would mention about '|||' in the definition since it's not normal to have these in instances.

Palipoor:
Fixed! Thanks.

danyaljj:
Thanks!

PR #267, opened 2021-09-15
Task 522 News Editorials

Palipoor:
From paper https://aclanthology.org/2020.coling-main.470.pdf

(I made a mistake and created this branch from another branch instead of the master, so I have commits from other tasks here. sorry)


danyaljj:
@gkaramanolakis wondering if you can help review this PR?

gkaramanolakis:
This PR contains two tasks: 521 and 522. I see that task 521 is reviewed separately at #263, so I will review task 522. 

Task 522 looks good but quite hard, so below are a few comments attempting to make it a bit easier.

In the task definition, could you copy the requirements for the summaries as they are stated in the paper? i.e., replace "concise, self-contained, persuasive, and useful for others" with the following: "thesis-indicative, persuasive, reasonable, concise, and self-contained" (except if I am missing something from the task definition/instructions). 

The paper provides more instructions for the annotators, which I believe should be included in the task definition, otherwise, the task is under-constrained and hard to address automatically. Here is an example of additional instructions mentioned in the paper: "Summary length was limited to be 20% of an editorial’s segments. In many editorials, the thesis may not be explicitly stated but rather implied by the author. For this reason, we allowed up to two segments to be labeled as thesis"

Also, your task definition mentions that "Your answer should consist of segments of text, in the exact order they appear in the main text, separated by comma" but I didn't find any non-consecutive segments of text separated by comma. (I checked all outputs automatically.) Is there any example I am missing where multiple segments are separated by comma? If not, then you could remove the corresponding part from the task description.

Do you think it makes sense to remove the "justification" part of the output summary and keep just the "thesis" part? If not, then I would suggest adding a few more negative examples to indicate the expected output format, e.g., an output without a "justification" field and an explanation mentioning that the output is wrong. 

Alternatively, could "justification" and "thesis" be merged into a single output? Given both "justification" and "thesis" fields, we should construct a special evaluation function just for this task, as generation metrics like BLUE cannot be directly applied to the outputs.

Hope this helps, and pls lmk if anything is unclear.

Palipoor:
@gkaramanolakis Thanks for the review! I will update the task.

Palipoor:
So I did what you suggested, and removed the justification, as it didn't have a good agreement between annotators. Can you check the task again? Thanks.

gkaramanolakis:
> So I did what you suggested, and removed the justification, as it didn't have a good agreement between annotators. Can you check the task again? Thanks.

Thanks a lot for the changes -- the task looks much better now! I have a few more comments: 

- All instances start with "thesis:" but the second negative example does not start with "thesis:". To be consistent you could either add "thesis:" as a prefix to the second negative example or remove "thesis:" from the rest. (Since the current task is to  summarize just the thesis, removing "thesis:" from all outputs would not be a problem.)
- There exist several instances with multiple outputs (e.g., [this one](https://github.com/allenai/natural-instructions-expansion/blob/c2c3bdb3e078f570562c2388c86f10d38e403588/tasks/task522_news_editorial_summary.json#L53)), where each output summary corresponds to a different annotator. A model trained for this task should clearly generate a single output summary. @danyaljj @Palipoor do you think we should keep all outputs in the file (that should require a special evaluation procedure for the given task) or should we find a way to select a single output?

Besides these two comments, everything else looks great to me.

Palipoor:
So I removed the "thesis:" from outputs, but for the second point, I don't know which ones I should keep because all of them have the maximum IAA score.

gkaramanolakis:
> So I removed the "thesis:" from outputs, but for the second point, I don't know which ones I should keep because all of them have the maximum IAA score.

Thanks for addressing the first comment! Regarding the second comment, I am also not sure about what selection criterion to use and couldn't find any paper that uses this dataset to train models. Everything else looks good to me, so if @danyaljj agrees we can commit the task as is. Thanks!

danyaljj:
> > So I removed the "thesis:" from outputs, but for the second point, I don't know which ones I should keep because all of them have the maximum IAA score.
> 
> Thanks for addressing the first comment! Regarding the second comment, I am also not sure about what selection criterion to use and couldn't find any paper that uses this dataset to train models. Everything else looks good to me, so if @danyaljj agrees we can commit the task as is. Thanks!

Yeah, it is fine. 

One more:  for `task521_trivia_question_classification`, clarify whether we can have more than one category for one input or not.

Palipoor:
@danyaljj Sorry it was my mistake to include the file for task 521 in this PR. I updated the task in its original PR #263  and merged it onto this branch. So I think it'll suffice to just merge this PR into master.

danyaljj:
Looks good to me! Leaving it for @swarooprm to review before merging.

swarooprm (approved):
LGTM.
I was wondering if it is feasible to create tasks to identify if the news summary matches the news article or select the appropriate matching article from the news summary, but not sure if it makes sense for the data you have.

PR #272, opened 2021-09-16
Task 524-529: ParsiNLU-sentiment-analysis

yeganehkordi:


danyaljj:
@amirrezamirzaei wondering if you can help with reviewing this PR?

amirrezamirzaei:
- in task 524-525 explanation states `You have to answer the question based on the review.` which isn't what we are doing in this task. 
- task 526-529 looks good.
- one thing about the classification labels, I don't understand the difference between `negative` and `very negative` (same for positive), I think it's better if we remove the second label entirely or explicitly state what the difference between them is.

yeganehkordi:
Thanks for reviewing!
In task 524-525, there is a question about one aspect and we have to categorize the sentiment toward the required aspect. Actually, we classify the answer to the question in this task. 
For example: `"من مزه موزی که میداد و بوی موزی که میداد رو دوست داشتم<sep>Question: نظر شما در مورد عطر، بو، و طعم این بیسکویت و ویفر چیست؟"`
I used the `negative` and `very negative` based on the original dataset. `Very negative` and `very positive` refer to extreme sentiment toward a product. I'll explain the difference in the definition now.

amirrezamirzaei:
thank you for the changes. I still think the definition for 524-525 isn't complete because we aren't answering any questions we are just classifying sentiments. can you change it to something like this `... question about the reviewer's sentiment toward one aspect of the food. You must infer the answer from the review and classify the reviewer sentiment.` 
everything else looks good to me.

yeganehkordi:
Thanks! It's done.

tasks/task524_parsinlu_food_aspect_classification.json
```diff
@@ -11,7 +11,7 @@
...
+    "Definition": "In this task, you will be given a food review and a question about the reviewer's sentiment toward one aspect of the food. You have to infer the answer to the question from the review. Classify the reviewer's sentiment into: \"no sentiment expressed\", \"very negative\", \"negative\", \"neutral\", \"positive\",  \"very positive\", and \"mixed\". The mixed category indicates reviews where none of the sentiments are dominant (mix of positive and negative, or borderline cases); hence it is hard to detect the primary sentiment. Also, assign neutral label to reviews that express no clear sentiment toward an entity or any aspect of it. The \"no sentiment expressed\" label should be assigned to the reviews where the given aspect didn't discuss in the text. \"very negative\" and \"very positive\" labels indicate extreme sentiment toward an aspect.",
```
amirrezamirzaei, line 14: isn't using `... question from the review and Classify ..` better?

yeganehkordi: By combining them, it's going to be a really long sentence. If you think it's better, I can fix it now.

amirrezamirzaei: yes your right but I still think the definition isn't complete . can we change it into something like `You have to infer the answer to the question from the review and classify it. Classify the reviewer's sentiment into...`

yeganehkordi: Sure! I updated them.

amirrezamirzaei:
thank you everything looks good to me.

danyaljj:
Thank you, all! 

@yeganehkordi I am a bit worried that the border between "x" and "very x" might be fuzzy and hence, average humans might not be able to clearly distinguish them. What do you think if we:
 - merged "negative" + "very negative" -> "negative",
 - merged "positive" + "very positive" -> "positive",
to reduce any potential subjectivity/noise?

yeganehkordi:
Sure! I'll merge them now.

danyaljj:
Thanks!

PR #275, opened 2021-09-17
Task 534 : farstail entailment

amirrezamirzaei:
Task 534 using [FarsTail](https://arxiv.org/abs/2009.08820)
I used the same definition as task 190.
not really sure about the output language here, Because it's a classification task.


danyaljj:
@yeganehkordi wondering if you can review this PR?

yeganehkordi:
Sure! I'll check it now.

tasks/README.md
```diff
@@ -372,3 +372,4 @@ Name | Summary | Category
 `task518_emo_different_dialogue_emotions` | Given different dialogue determine if they have the same emotion | Classification, Sentiment Analysis
 `task519_aquamuse_question_generation` | Given an answer generate a question that would be answered by the answer given | Question Generation
 `task520_aquamuse_answer_given_in_passage` | Given a passage and a question determine if the question can be answered by the passage | Classification
+`task534_farstail_entailment` | Given two sentences in persian choose whether they agree/disagree/neither with each other. | Classification
```
yeganehkordi: Please change `in persian choose` to `in Persian, choose`. And use a comma instead of "/".

yeganehkordi (commented):
Thanks! Some minor comments:
Would you please add the "instruction_language" field referring to the language of definition? Also, please move language fields before the definition for consistency across the tasks that used these fields.
I think you can safely drop ` in Persian Language written in persian alphabet` because we already have the "input_language" field. 
Please add a space after the dot in `.Your job`. Also, I think using a comma instead of "/" in `(entailment)/disagree` makes more sense. 
Consider removing comma in `with each other, or if` and adding a comma before respectively: `E, C, and N respectively`.
Please remove the extra dot from the third example and start explanations with a capital letter.

amirrezamirzaei:
changed part of the definition entirely.

tasks/README.md
```diff
@@ -372,4 +372,4 @@ Name | Summary | Category
 `task518_emo_different_dialogue_emotions` | Given different dialogue determine if they have the same emotion | Classification, Sentiment Analysis
 `task519_aquamuse_question_generation` | Given an answer generate a question that would be answered by the answer given | Question Generation
 `task520_aquamuse_answer_given_in_passage` | Given a passage and a question determine if the question can be answered by the passage | Classification
-`task534_farstail_entailment` | Given two sentences in persian choose whether they agree/disagree/neither with each other. | Classification
+`task534_farstail_entailment` | Given two sentences in Persian, choose whether they agree,disagree,neither with each other. | Classification
```
yeganehkordi: Could you please add a space after the comma?

tasks/task534_farstail_entailment.json
```diff
@@ -8,31 +8,32 @@
     "Categories": [
         "Classification"
     ],
-    "Definition": "In this task, you're given a pair of sentences in Persian Language written in persian alphabet.Your job is to choose whether the two sentences clearly agree (entailment)/disagree (contradiction) with each other, or if this cannot be determined (neutral). Your answer must be in the form of the letters E, C, and N respectively. The sentences have been separated by a new line character.",
     "Input_language": "fa",
     "Output_language": "en",
+    "Instruction_language" : "en",
+    "Definition": "In this task, you're given a pair of sentences in Persian Language written in persian alphabet. Your job is to choose whether the two sentences agree (entailment), disagree (contradiction) or neither (neutral). Your answer must be in the form of the letters E, C, and N respectively. The sentences have been separated by a new line character.",
```
yeganehkordi: I think it's better to drop `in Persian Language written in persian alphabet.`, because we already know that input language is Persian.
Please add a comma in `(contradiction) or neither` after (contradiction) and in ` N respectively` after "N".
It will be more clear if you explain which word each letter is referring to.

amirrezamirzaei: I think we should specify the alphabet we are using because some translation tasks use the native alphabet and the English alphabet ([example](https://github.com/swarooprm/CSE576HW/blob/69ed0eabf66355f63d97b8d93301d7b5aa72c582/tasks/Task361_tamil_offenseval_dravidian_classification.json)). but I will remove it if you think I should.
also is comma after "N" grammatically correct?

yeganehkordi: This makes sense. So please replace it with `in the Persian Language written in the Persian alphabet`.
Yes, As far as I know, we need a comma before respectively, when it comes at the end of the sentence.

danyaljj:
@yeganehkordi  check if you're happy with the changes?

yeganehkordi:
It looks good!
To be picky, I think `in Persian Language` can be replaced with `in the Persian Language`, and I guess there is an extra space in 
 the `newline character`.
Please merge the task if these comments are too minor :)

danyaljj:
Thanks! Merging!

PR #296, opened 2021-09-21
Tasks 747-749 and 614 from Glucose

Palipoor:
Hi, I created the first task and I will create a few more from this dataset (it has the potential for at least 10 tasks actually)
paper: https://arxiv.org/pdf/2009.07758.pdf

It'd be great if someone could review this, because I will create the other tasks in the same manner.


danyaljj:
@cosmicishan  wondering if you can review this PR?

cosmicishan:
> @cosmicishan wondering if you can review this PR?

Sure I will review it.

Mirzyaaliii:
Just a quick suggestion, @Palipoor number **672** is already taken in #307.
So, you need to change the number for your task.

Palipoor:
> Just a quick suggestion, @Palipoor number **672** is already taken in #307.
> So, you need to change the number for your task.

Thanks! Fixed it.

cosmicishan:
- General comments 

1. Remove the full stop at the end of README.
2. It's better if you only keep one element in the source list and also provide a link to the Dataset as well.
3. And I am also concerned about the categories of all these tasks where you addressed it as 'Answer Generation'. I can't confidently say it's a wrong category. But in my opinion, I think it will probably come under the category: 'Text Generation' or 'Sentence Generation'. Maybe you should go through this with @swarooprm or @danyaljj since I have a doubt about the label of the category but am not sure which one is correct.

- Task 614

```
        {
            "input": "story: My friend R and her husband G were going to a beach party. They were bringing cups. G got bad news from work and did not want to go to the party. R asked who will bring the cups. G said they will manage, and they did not go to the party.\n selected sentence: G said they will manage, and they did not go to the party.",
            "output": [
                "R and G don't want togotothe party >Causes/Enables> R and G don't go tothe party"
            ]
        },

```
I have one concern in the output of these instances. Why there's no space between 'togotothe' and 'tothe'? And there are many other examples like this.

- Task 673

1. Edit the definition here where you mistakenly have added 'Y' in between.

> Y You should write your answer in the form \" A >Motivates> B\".

- Task 674

Same as Task 614 there are the same mistakes in the output of instances where there's no space between the words. Like here in this instance multiple words combined into one word like: 'Jamieclimbedupthe'. And there are so many instances like this.

```
        {
            "input": "story: Jamie ran outside to play in the snow. She climbed up the slide in her backyard. She fell off the slide onto the ground! Jamie started crying loudly because her elbow was bruised. She went back inside to have hot chocolate.\n selected sentence: She climbed up the slide in her backyard.",
            "output": [
                "Jamieclimbedupthe slideinher backyard>Causes/Enables> Jamiefelloffthe slideontothe ground "
            ]
        },
        {
```

- Task 675

1. Edit the definition here where you mistakenly have added 'Y' in between.

> Y You should write your answer in the form \" A >Causes> B\".

2. My other concern from this Task is that you clarify that it must choose the verb form between 'feel(s), want(s), or like(s)' but in Instances, there's no verb form besides 'feel(s)' so fix the Instances by making it balanced across all labels or change the definition.


Besides this, it all looks great to me. It's all really a very creative and unique task.

swarooprm:
@Palipoor, when you fix the changes, can you also rename your tasks to task747-749 (sorry I should have seen it earlier and asked others to rename their PR)? I hope this does not require a lot of effort for you, since you will be changing files to address reviewer coment.

Palipoor:
Sure! Will do tomorrow.

Palipoor:
@cosmicishan Thanks for the thorough review! The spacing problem was because of a bug I had in preprocessing the outputs. I fixed it. Also fixed other issues and changed the category to sentence generation.

cosmicishan:
> @cosmicishan Thanks for the thorough review! The spacing problem was because of a bug I had in preprocessing the outputs. I fixed it. Also fixed other issues and changed the category to sentence generation.

Seems like Task 614 didn't get updates. Because there's still no space between words, source list have 2 elements and no paper link, the caegory is 'Answer Generation' and there's fullstop at the end of it's README.

Also you should specify which link is from dataset and which link is from paper.

For example:

`GLUCOSE (Dataset  - https://github.com/ElementalCognition/glucose/)  (Paper - https://arxiv.org/pdf/2009.07758.pdf)`

And I am also very confused about the difference between Task 614 and Task 748. Both of them looks very similar to me. If the goal in Task 748 is use the exact sentence given in the input and use it before >Causes/Enables> then I think it's correctly expressed in Positive examples but not in Instances. I don't know if I am misunderstading something here or missing something out here.

Palipoor:
@cosmicishan I checked task 748 and the instances seemed correct to me. 
I fixed other problems.

cosmicishan:
@swarooprm I have done a review from my side and it looks good to me. I only have one concern, which is I couldn't find the difference between Task 614 and Task 748. So please do the final check from your side.

danyaljj:
>  @swarooprm I have done a review from my side and it looks good to me. I only have one concern, which is I couldn't find the difference between Task 614 and Task 748. So please do the final check from your side. 

Here is the diff between the two tasks: 
Screen Shot 2021-09-29 at 2 17 04 PM

I think the difference is "before" vs "after".

swarooprm (approved):
LGTM

PR #297, opened 2021-09-21
Task 615 Movieqa

cosmicishan:
task615_moviesqa_answer_generation -> Given a question from an open movie database, generate an answer for that. 

Thanks in advance!!

Please review.


danyaljj:
Looks good to me! Leaving it here for @swarooprm to double-check before merging.

swarooprm:
Looks good, but my concern is that for a given question, there exist multiple correct answers, however only one is provided in the output. If you have access to multiple correct answers, add them as list in instance outputs, otherwise we can still accept this task, but we have to make sure to keep it in the training split. @cosmicishan

cosmicishan:
> Looks good, but my concern is that for a given question, there exist multiple correct answers, however only one is provided in the output. If you have access to multiple correct answers, add them as list in instance outputs, otherwise we can still accept this task, but we have to make sure to keep it in the training split. @cosmicishan

In the dataset, only one answer is given if multiple answers was available I would have definitely added it. You can check the demo of the dataset in the hugging face link I provided.

swarooprm (approved):
LGTM! Merging.

For future: We can have jeopardy type tasks, where given an answer model needs to generate a question. We don't have such tasks. You might be able to create such a task from this MovieQA dataset.

PR #298, opened 2021-09-21
Task 616 CoLA

cosmicishan:
task616_cola_classification -> Given a sentence you have to return if it is acceptable or unacceptable. 

Thanks in advance!!

Please review.

### DIFF
--- a/tasks/task616_cola_classification.json
+++ b/tasks/task616_cola_classification.json
@@ -8,7 +8,7 @@
     "Categories": [
         "Classification"
     ],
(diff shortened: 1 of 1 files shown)


danyaljj:
You need to define "acceptability" in what sense (which seems to be in terms of whether the given sentence makes sense or not).

cosmicishan:
> You need to define "acceptability" in what sense (which seems to be in terms of whether the given sentence makes sense or not).

I fixed that. Can you review it again?

danyaljj:
Looks good to me. Leaving it here for @swarooprm to review it once more before merging.

swarooprm (approved):
LGTM.
I assume classes are balanced.
One new task can be given a grammatical incorrect sentence, correct it, but I don't know if the COLA dataset contain such instances.

PR #299, opened 2021-09-21
Task 617-618 Amazon Review

cosmicishan:
task617_amazonreview_category_text_generation -> Given an Amazon product review your task is to generate the category of product. 
task618_amazonreview_summary_text_generation -> Given an Amazon product review your task is to generate the summary of the review. 

Thanks in advance!!

Please review.


danyaljj:
Looks good to me! Leaving it here for @swarooprm to double-check before merging it.

swarooprm (approved):
LGTM.
Re: more task suggestions- you can create classification tasks that check if the sentence and its category match, similarly for the summary. You can make the task even harder by making it a column matching task similar to the ones we had in high school. column1-contain many reviews, column 2- categories or summaries, the goal is to generate matches e.g  a valid answer could be c11-c23, c12-c21, c31-c22. @cosmicishan

PR #303, opened 2021-09-22
Tasks 629-635: dbpedia_14 and allegro_reviews datasets

divya0627:


swarooprm:
Reviews: https://github.com/swarooprm/CSE576HW/pull/87

danyaljj:
If the question is always the same ("Question: Is the text about a person?"), why do you need to repeat it for each instance?

divya0627:
> If the question is always the same ("Question: Is the text about a person?"), why do you need to repeat it for each instance?

@danyaljj You are right, I can change it to a classification task instead of the answer generation task.

divya0627:
> > If the question is always the same ("Question: Is the text about a person?"), why do you need to repeat it for each instance?
> 
> @danyaljj You are right, I can change it to a classification task instead of the answer generation task.

@danyaljj  Made changes to the task as per the review.

danyaljj:
Thanks! Another suggestion: for the tasks that are not English, add a language tag, as per #113.

divya0627:
> Thanks! Another suggestion: for the tasks that are not English, add a language tag, as per #113.

@danyaljj Made changes to tasks 634,635 as per the review

PR #176, opened 2021-09-04
Task 309-311: RACE

Mirzyaaliii:
task309_RACE_answer_generation  -->  Given an article, a question and four options; provide the correct answer for the question based on the article.
task310_RACE_classification  -->  Given an article, a question and four options; provide the incorrect answer for the question based on the article.
task311_RACE_question_generation  -->  Generate a question based on the given article and an answer.

Please review.

PR #309, opened 2021-09-23
Task - 649: RACE

Mirzyaaliii:
One more task from the RACE dataset. (As suggested in PR #176)

task649_race_blank_question_generation -> Generate a fill-in-the-blank question based on the given article and an answer.

Please review.


swarooprm:
This task looks good to me. One question: If you have multiple question corresponding to an input, make sure to add them as a list in instance output.

Also, there are 3 old files which you are pushing here. Those files are already part of our tasks. This is intentional? If not, can you remove those files here?

Mirzyaaliii:
> This task looks good to me. One question: If you have multiple question corresponding to an input, make sure to add them as a list in instance output.
> 
> Also, there are 3 old files which you are pushing here. Those files are already part of our tasks. This is intentional? If not, can you remove those files here?

There is only one question per input because there exist 1 question for the given _answer_ and _article_.
And thank you for pointing out extra files, I removed those files.

swarooprm (approved):
LGTM

PR #312, opened 2021-09-24
Task 669-671: AmbigQA

Mirzyaaliii:
task669_ambigqa_answer_generation -> Given an open-domain, provide an answer for that question.
task670_ambigqa_question_generation -> Given an ambiguous question, return all possible questions which clarify the given question.
task671_ambigqa_text_generation -> Given an ambiguous question, provide one clarifying question and answer for the generated question.

Please review.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -517,14 +517,6 @@ Name | Summary | Category
 `task532_europarl_en-es_classification` | Given a English sentence and its corresponding Spanish translation, classify whether it is correct or not. | Classification
 `task533_europarl_es-en_language_identification` | Given a sentence, identify  whether it is in English or Spanish. | Language Identification
 `task534_farstail_entailment` | Given two sentences in Persian, choose whether they agree, disagree, neither with each other. | Classification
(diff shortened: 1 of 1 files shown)


swarooprm:
@nrjvarshney, wondering if you can help review this tasks?

swarooprm:
@yeganehkordi, wondering if you can help review these tasks.

yeganehkordi (commented):
Thanks! Looks good to me.
I think it's better to say, `Your task is to return a question` in task 670. All the possible questions might be unclear. Although this task has many correct answers here, we only need one answer, and we can have all the possible answers in output.

swarooprm:
@Mirzyaaliii, these tasks look good to me as well. I agree with @yeganehkordi's comment: we should ask the model to generate only one question (instead of all possible questions). Just that minor fix and this PR is good to go!

Mirzyaaliii:
Thanks for the comments!!

Changes are done as per suggestion.

danyaljj:
Thanks! Sorted several lines in the tasks readme and merging the PR after the tests turn green.

PR #320, opened 2021-09-26
Task 761-765: app_review and emea 

vijaykumawat256:


swarooprm:
Reviews: https://github.com/swarooprm/CSE576HW/pull/33

tasks/task764_emea_bg_el_classification.json
```diff
@@ -0,0 +1,1061 @@
+{
+    "Contributors": [
+        "Vijay Kumawat"
+    ],
+    "Source": [
+        "emea bg-el (dataset link- https://huggingface.co/datasets/emea)"
+    ],
+    "Categories": [
+        "Classification"
```
danyaljj, line 9: These are clearly "translation" tasks; right?

vijaykumawat256: Task 765 is a translation task for the same task764 In this task, I have taken both languages sentences and compared whether the generated language is the correct translation or not from the dataset. If it's correct then I have generated output as True label, otherwise False label. So I have categorised it as a Classification Task.

danyaljj: Got it, thanks!

PR #329, opened 2021-09-27
Tasks 850-852: Synthetic data

kurbster:
For the longest palindrome task, I randomly choose 8 characters for each example and generated a string from those 8 characters. This helped increase the length of the resulting palindromes.


danyaljj:
Thanks! Overall looks good! 
Here are some suggestions: 

 * For the palindrome task, I worry if humans can answer the instances, given that some are super-long: 
```
{
            "input": "nnqnqvvvvvwqnqnnvvwggwnggvwqgwqvnvnwnvnqvqqqvnvqnvqvvvvqqqvqqgwqvwnvnvnvqvqgvqggwvvqnvqvvnvvqgwwnqwvqgqvqngqvwqgvgwwwqvvwgnwqwqqnnqnnqvqvnwnwgwwvqvnggnqvngqn",
            "output": [
                "nvnwnvn"
            ]
        },
```
wondering if we can limit ourselves to shorter input strings? (say, to max = 15 characters) 

* For multiply even / odd tasks, do you need to state "Zero should count as an even integer.", given that 2 x 0 = 0.

kurbster:
Sorry for the late reply I have been sick all week. 

For the palindrome task, I used shorter sentences but then the average palindrome length was 3 and the longest was 5. So I made the strings long to make the palindromes longer. Is it ok if every palindrome is short? Should I only use 2 characters (chosen at random) for each string then I could force long palindromes with short strings (10-15 chars)?

For the even and odd tasks, I wanted to specify that zero is even so it would be multiplied in the output. If the model thought zero was odd it would not multiply and the answer would be nonzero. That being said should I remove this because it should be known that zero is even?

danyaljj:
> For the palindrome task, I used shorter sentences but then the average palindrome length was 3 and the longest was 5. So I made the strings long to make the palindromes longer. Is it ok if every palindrome is short? Should I only use 2 characters (chosen at random) for each string then I could force long palindromes with short strings (10-15 chars)?

Ya makes sense. 

> For the even and odd tasks, I wanted to specify that zero is even so it would be multiplied in the output. If the model thought zero was odd it would not multiply and the answer would be nonzero. That being said should I remove this because it should be known that zero is even?

Yeah, I think so.

kurbster:
Sounds good, I will update the tasks today. One last question, for the palindrome task if I change the instances to a 2 character alphabet, should I change my positive and negative examples as well? For example, instead of using `racecar` as an example, I would use `errerre`.

danyaljj:
> Sounds good, I will update the tasks today. One last question, for the palindrome task if I change the instances to a 2 character alphabet, should I change my positive and negative examples as well? For example, instead of using `racecar` as an example, I would use `errerre`.

Yeah, I think we want examples to be compatible with the instances + task definition.

danyaljj:
Looks good, thanks!

PR #338, opened 2021-09-28
Tasks 890-893: GCWD and gap

selvagmj:


tasks/task890_gcwd_classification.json
```diff
@@ -0,0 +1,1237 @@
...
+        "Selvaganesh Muthu"
+    ],
+    "Source": [
+        "GWSD",
+        "https://github.com/yiweiluo/GWStance"
+    ],
+    "Categories": [
+        "Classification"
+    ],
+    "Definition": "Read the passage and find if the passage agrees, disagrees, or has a neutral stance on whether Global warming is caused by human activities. Answer only with keyword (a) agrees - if passage agrees with the target (b) disagrees - if passage disagrees with the target (c) neutral - if passage neither agrees nor disagrees with the target.",
```
danyaljj, line 12: A bit verbose, but reads well to me.

tasks/task891_gap_coreference_resolution.json
```diff
@@ -0,0 +1,1232 @@
...
+        "Selvaganesh Muthu"
+    ],
+    "Source": [
+        "gap",
+        "https://huggingface.co/datasets/gap"
+    ],
+    "Categories": [
+        "Coreference Resolution"
+    ],
+    "Definition": "Read the passage and find the corresponding person for the given pronoun. The pronoun-offset is the character position of first character of the pronoun from start.",
```
danyaljj: Instead of "chracter offset", which is a bit difficult for humans to read/comprehend, how about we indicate the target pronoun within the sentence?  For example, we can indicate it as `.... [he] ...` or  `**he**`. 
What do you think?

selvagmj: Yeah, we can indicate it that way. But in certain rare cases, the text itself may contain such symbols. Will that be a problem? Shall I change it?

danyaljj: Yeah, please check first if the texts contain such symbols. You can always come up with an open-close sequence that is not contained in the text.

aarunku5:
@selvagmj can you update your tasks based on the comments? Thanks!

selvagmj:
@aarunku5 I update the tasks by EOD tomorrow.

danyaljj:
Thanks!

PR #339, opened 2021-09-28
Task 894- 901: miam, freebase_qa

karthikmuru:
https://github.com/swarooprm/CSE576HW/pull/114

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -853,7 +853,7 @@ Name | Summary | Category | Domain | Input Language | Output Language
 `task891_gap_coreference_resolution` | Finding names for given pronouns | Coreference Resolution
 `task892_gap_reverse_coreference_resolution` | Finding pronouns for given names | Reverse Coreference Resolution
 `task893_gap_fill_the_blank_coreference_resolution` | Fill the blanks with corresponding pronouns | Coreference Resolution
-`task894_miam_dialogue_act_classification` | Classify Dialogue utterances to Dialogue acts | Classification
+`task896_miam_language_classification` | Classify the language for Dialogue utterances | Classification
 `task897_freebase_qa_topic_question_generation` | Generate question for the given topic | Question Generation 
 `task898_freebase_qa_answer_generation` | Answering questions based on an open ended topic | Answer Generation
 `task899_freebase_qa_topic_generation` | Generate the specific topic for a given question | Topic Generation


danyaljj:
* In the definition you mentioned `a utterance from a dialogue sequence between 2 characters 'g' and 'f'`. What confuses me is that, in certain inputs, I don't see any mention of "f" or "g": 
```
"input": "Utterance: 'okay the start part is at the top left-hand corner'. Dialogue: ''",
```
 * Should `'reply_n' (Yes-No-Question)` this `'reply_n' (reply to Yes-No-Question)` or `'reply_n' (Yes-No-Reply)`? 
 * I would drop the mention of the URL: `Detailed explanation of the categories is given in this link (https://groups.inf.ed.ac.uk/maptask/interface/expl.html)` The definition should be self-contained.

karthikmuru:
> * In the definition you mentioned `a utterance from a dialogue sequence between 2 characters 'g' and 'f'`. What confuses me is that, in certain inputs, I don't see any mention of "f" or "g":
> 
> ```
> "input": "Utterance: 'okay the start part is at the top left-hand corner'. Dialogue: ''",
> ```
> 
> * Should `'reply_n' (Yes-No-Question)` this `'reply_n' (reply to Yes-No-Question)` or `'reply_n' (Yes-No-Reply)`?
> * I would drop the mention of the URL: `Detailed explanation of the categories is given in this link (https://groups.inf.ed.ac.uk/maptask/interface/expl.html)` The definition should be self-contained.

Hi @danyaljj. Thanks for the review! I have addressed the comments
1. When the 'Dialogue' param is empty, then the dialogue is just beginning with that utterance and no one has spoken before that. Added this instruction to the definition.
2. Have elaborated on the meaning of each class. Changed the explanation for 'reply_n' also.
3. Removed the external link and added the explanation to the description.

danyaljj:
> When the 'Dialogue' param is empty, then the dialogue is just beginning with that utterance and no one has spoken before that. Added this instruction to the definition. 

I see. A basic question: what is the difference between an "utterances" and "dialogue"?

aarunku5:
> > When the 'Dialogue' param is empty, then the dialogue is just beginning with that utterance and no one has spoken before that. Added this instruction to the definition.
> 
> I see. A basic question: what is the difference between an "utterances" and "dialogue"?

@karthikmuru can you address this? Thanks!

karthikmuru:
Sorry for the late reply! 
An utterance is a single statement from a person, while a dialogue consists of multiple utterances. @danyaljj 

> > > When the 'Dialogue' param is empty, then the dialogue is just beginning with that utterance and no one has spoken before that. Added this instruction to the definition.
> > 
> > 
> > I see. A basic question: what is the difference between an "utterances" and "dialogue"?
> 
> @karthikmuru can you address this? Thanks!

danyaljj:

aarunku5:
In 894, can you update your examples instances so that the speaker (f/g) tag appears in the utterance as well? Define explicitly that the dialog is comprised of several utterances too as you've explained above.
Also for this part-- Categories:['instruct-- ....]' -- can you rephrase that to remove that [] notation? Make it something like possible answer categories are: 1. instruct---, 2.--- etc.

895- differentiate dialog and utterance in the definition, 

898, 899- add details  in the definition about rules present for generating the answer (if any)

For all your tasks, can you add the following fields as well (change languages if applicable):

"Input_language": [
        "English"
    ],
    "Output_language": [
        "English"
    ],
    "Instruction_language": [
        "English"
    ]

danyaljj:
- Dropped tasks 895 and 896 in the interest of time. We can bring them back next in future PRs, if we improve their definitions.

PR #363, opened 2021-10-02
Tasks 1087-1089 and 1146-1147 - Synthetic Data - Answer Generation

Ravsehajsinghpuri:
Task Description - Given a list of integers and a target sum, return a pair of integers from the list which sum to the target sum.


swarooprm:
@Ravsehajsinghpuri: Thanks for the PR. It looks good overall.
I have fixed some minor things in the definition and readme. The idea is to not to use maths and yet explain the task. 
Some comments:
- Seems like labels are already balanced. Please check and shuffle instances. 
- Also, try to have examples from both 1 and 2 in positive examples.
- Task 1087: there can be multiple answers possible for a given target sum. For positive and negative examples its fine to have single output, but for instances we need to have all possible combinations so that evaluation is fair. Every answer in reverse needs to be added in the output of instances along with other possible answers.
E.g. ["[1,2]",[2,1],"[3,0]","[0,3]"] should be the output for input: [0,2,1,3,4] and target:3

Ravsehajsinghpuri:
> @Ravsehajsinghpuri: Thanks for the PR. It looks good overall. I have fixed some minor things in the definition and readme. The idea is to not to use maths and yet explain the task. Some comments:
> 
> * Seems like labels are already balanced. Please check and shuffle instances.
> * Also, try to have examples from both 1 and 2 in positive examples.
> * Task 1087: there can be multiple answers possible for a given target sum. For positive and negative examples its fine to have single output, but for instances we need to have all possible combinations so that evaluation is fair. Every answer in reverse needs to be added in the output of instances along with other possible answers.
>   E.g. ["[1,2]",[2,1],"[3,0]","[0,3]"] should be the output for input: [0,2,1,3,4] and target:3

swarooprm:
Task 1087: In instance ,
"output": [
                "['[19, 37]', '[37, 19]', '[24, 32]', '[32, 24]']"
            ]
should be 
"output": [
                "[19, 37]", "[37, 19]", "[24, 32]", "[32, 24]"
            ]
Similarly, change for all.
Task 1088: looks good, just that in examples, explanation should be before output (this is the trend in other tasks), in the explanation just mention that 'i' is the position of element in an array, otherwise it is undefined at this point
Task 1089: explanation should be before output (this is the trend in other tasks)
Task 1090 and 1091: Rename to 1146 and 1147 (as other task names are taken), add source weblink in the source field, explanation field should be before output.

This PR is good to go once the above issues get fixed. @Ravsehajsinghpuri

Ravsehajsinghpuri:
Task 1087: In instance ,
"output": [
"['[19, 37]', '[37, 19]', '[24, 32]', '[32, 24]']"
]
should be
"output": [
"[19, 37]", "[37, 19]", "[24, 32]", "[32, 24]"
]

For this, I would suggest, "output" :[
["[19, 37]", "[37, 19]", "[24, 32]", "[32, 24]"]
]
As there will be instances where it is the following:
"output": [
]
This shows an error when checks are processed as len(output)>0

Ravsehajsinghpuri:
Have made the required changes, please take a look and merge

swarooprm (approved):
LGTM.

PR #370, opened 2021-10-04
Tasks 1148-1153

Ravsehajsinghpuri:


danyaljj:
@cosmicishan  wondering if you can review this PR?

cosmicishan:
> @cosmicishan wondering if you can review this PR?

Sure I will review it.

cosmicishan:
Can you add those tasks so I could review them?

Ravsehajsinghpuri:
@cosmicishan which tasks? I have already added 6 tasks to this PR

cosmicishan:
> @cosmicishan which tasks? I have already added 6 tasks to this PR

Got it. I will review it now.

Ravsehajsinghpuri:
@cosmicishan Thanks!

cosmicishan:
* General Comments 

1. Correct grammar and typos in the explanation and definition from all the tasks. Few suggestions: Capitalise the first word and add a full stop in the end.
2.  In the source list if you have access to the link of the paper do provide that. And it would be better if you mention the link is from a dataset or a paper and also add the name of the dataset before it.

For example

`"Name of the dataset" (Dataset: https://data.world/samayo/country-capital-city/workspace/file?filename=country-by-capital-city.json)`

* Task 1146, 1147

In definition mention that the dataset is open-domain and in input, you are given a country name that can be answered based on factual information.

* Task 1149 

It's a classification task but you add 'answer generation' in the category. So change that.

* Task 1150-1151-1152-1153

I wonder if that task is even necessary because the goal could easily be achieved through simple python code. We can easily generate a correct output of the input from all the tasks with simple python code.

@swarooprm @danyaljj Please check the last 4 tasks which I am not sure whether it's acceptable or not.

swarooprm:
@Ravsehajsinghpuri. these tasks look good, one minor suggestion:
In some tasks for example the ascii character task, in negative examples you don't say what would have been the correct answer. Do that wherever this is missing and this PR should be good to go.

@cosmicishan, it is ok to create tasks that can be solved using a simple python code because the whole point is to check if a combined model can solve everything provided in instruction without requiring to explicitly code

cosmicishan:
> @Ravsehajsinghpuri. these tasks look good, one minor suggestion: In some tasks for example the ascii character task, in negative examples you don't say what would have been the correct answer. Do that wherever this is missing and this PR should be good to go.
> 
> @cosmicishan, it is ok to create tasks that can be solved using a simple python code because the whole point is to check if a combined model can solve everything provided in instruction without requiring to explicitly code

That makes sense. Thanks for explaining.

PR #372, opened 2021-10-04
Task 1187 Politifact 

cosmicishan:
task1187_politifact_classification -> Given a statement and subject of discussion, your task is to classify whether it's a correct subject or not. 

Please review.


danyaljj:
Here as well, Amazon review files seem to be redundant addition.

cosmicishan:
> Here as well, Amazon review files seem to be redundant addition.

I removed it. I don't know how it got added.

Palipoor:
Hi,
It seems like your task number has changed. please reflect it in readme and file names.
Also, these comments:
- Here in the task definition: "of the discussion is the correct for the given statement" please remove the "the" before "correct". Also, maybe change " your task is to generate" to "your task is to determine".
- Please reword the explanation of the third positive example. You don't need to mention "Military". It'd suffice if you explain that election is not a correct subject.
- The explanation for the second negative example is a bit weird. 
- In the explanation for the third negative example, please change "It's only supposed to return True or False." to "You're only supposed to answer with true or false". 
- Also, please point out in the definition that the subject can be multiple words.

cosmicishan:
> Hi, It seems like your task number has changed. please reflect it in readme and file names. Also, these comments:
> 
> * Here in the task definition: "of the discussion is the correct for the given statement" please remove the "the" before "correct". Also, maybe change " your task is to generate" to "your task is to determine".
> * Please reword the explanation of the third positive example. You don't need to mention "Military". It'd suffice if you explain that election is not a correct subject.
> * The explanation for the second negative example is a bit weird.
> * In the explanation for the third negative example, please change "It's only supposed to return True or False." to "You're only supposed to answer with true or false".
> * Also, please point out in the definition that the subject can be multiple words.

Thank you for the thorough review and all the suggestions you provide to improve the definition and explanations. I have fixed that. Please review it again.

Palipoor (approved):

danyaljj:
The PR seems to be removing two tasks? 😮

cosmicishan:
> The PR seems to be removing two tasks? open_mouth

You told me in the above comment that some redundant files got added so I deleted them. So let me know what I should do if I should keep them or not?

danyaljj:
Looks like the two of your PRs (this and #299) shared their git history, which has caused this conflict here (which is not a good practice when opening multiple pull requests). Your easiest solution is to copy the additions here to a fresh checkout of the main branch and open a new PR.

Issue #283, opened 2021-09-20
Reply suggestion dataset 


danyaljj:
https://www.microsoft.com/en-us/research/uploads/prod/2021/06/mrs_acl2021_zhang.pdf

Palipoor:
I will take this.

PR #375, opened 2021-10-05
Add task 1160 from MRS

Palipoor:
This task is created from the MRS dataset from this issue #283.
However, I am in doubt whether this is a good addition or not. The data is driven from Reddit replies and they're not good quality examples to learn from. I've cleaned them as much as I could but there're still lots of nonsense going on. I'm submitting the English task, to get other people's opinions. If it's good enough, or there's a good way to filter out nonsense, I will go on and add other languages too.
@swarooprm @danyaljj


tasks/README.md
```diff
@@ -514,7 +514,7 @@ Name | Summary | Category
 `task608_sbic_sexual_offense_binary_classification` | Determine whether the post is sexually offensive/explicit or not | Binary Classification
 `task609_sbic_potentially_offense_binary_classification` | Determine whether the post is potentially offensive or not | Binary Classification
 `task610_conllpp_ner` | Recognize and label proper nouns (Named Entity Recognition) | NER/Labeling
-`task611_mutual_multi_turn_dialogue | Given a conversation between two people and 4 options on how the conversation should continue, choose the most reasonable option | Multiple-Choice Question
+`task611_mutual_multi_turn_dialogue` | Given a conversation between two people and 4 options on how the conversation should continue, choose the most reasonable option | Multiple-Choice Question
```
danyaljj, line 517: 👍

danyaljj:
Yeah, the data is quite noisy. I am leaning towards a "no", unless we can somehow clean it up around a particular subject. 
@swarooprm feel free to share your thoughts.

swarooprm:
I like this task, but I agree that noise is a concern.
I see that, longer instances (both longer inputs and outputs) are less prone to noise; I checked a few instances only though.
Check if this is true and we can filter instances based on that.

Palipoor:
I checked again and I think there's noisy data in short instances too.

swarooprm:
> I checked again and I think there's noisy data in short instances too.

What I meant above was to retain only the longer instances. Longer instances seem to contain lesser noise.
It's ok if you still see noise in longer sentences and in that case I am fine if we drop this task.
Subject wise filtering may be another option (some subjects may contain less noise e.g. scientific topic discussion)

Palipoor:
Oh sorry, I didn't read it carefully!
I will check it tomorrow and update the task if it's ok.

Palipoor:
Sorry for being late. I think it makes sense to keep the longer instances(not sure about the threshold though). Should I add other languages too?

swarooprm:
> Sorry for being late. I think it makes sense to keep the longer instances(not sure about the threshold though). Should I add other languages too?

Yes, if you have time, feel free to add. It's also fine if you skip this and decide to focus on other ToDos we have in this project.

danyaljj:
I agree with Swaroop. If cleaning up this PR will take more than 1hr, I would say, it's not worth it.

PR #379, opened 2021-10-06
Task 384-385 Socialiqa(replaced)

cosmicishan:
task384_socialiqa_question_classification -> You're given context, an answer, and question. Your task is to classify whether the question is correct or not. 
task385_socialiqa_incorrect_answer_generation -> You're given a context, a question, three options. Your task is to return an incorrect answer from the option. 

I have closed the previous PR which had noise in the task and replaced the numbers with new tasks. 

Please review.


swarooprm:
@cosmicishan How these are different from the SocialQA tasks you already created and are accepted?

cosmicishan:
> @cosmicishan How these are different from the SocialQA tasks you already created and are accepted?

I add the ones you suggested. Before closing the pull request you suggested tasks I could do from this dataset. So preciously I made a task to generate the question and here I made a task to check whether the given question is correct or not. And in the MCQ task I generate incorrect answers in this one.

aarunku5:
Thanks, a few corrections:

384:
Change defintion to:
In this task, you're given a context passage, an answer, and a question. Your task is to classify whether the question for this answer is correct or not, based on the given context with commonsense reasoning about social situations. If its correct ,return \"True\" else \"False\".
Can you split your example explanations into shorter sentences? There are also a couple of grammatical errors to be fixed.

385:
Change defintion to:
In this task, you're given a context passage, a question, and three answer options. Your task is to return an incorrect answer option to the question from the choices given. For all questions, only one of the three answer options is correct. Pick one of the two incorrect answer options as the output.
P1: So, C is the correct answer is repeated in the explanation
N1: which having an--> fix grammar
N2: would to be--> fix grammar
N2, N3: So C correct answer--> fix grammar

cosmicishan:
> Thanks, a few corrections:
> 
> 384: Change defintion to: In this task, you're given a context passage, an answer, and a question. Your task is to classify whether the question for this answer is correct or not, based on the given context with commonsense reasoning about social situations. If its correct ,return "True" else "False". Can you split your example explanations into shorter sentences? There are also a couple of grammatical errors to be fixed.
> 
> 385: Change defintion to: In this task, you're given a context passage, a question, and three answer options. Your task is to return an incorrect answer option to the question from the choices given. For all questions, only one of the three answer options is correct. Pick one of the two incorrect answer options as the output. P1: So, C is the correct answer is repeated in the explanation N1: which having an--> fix grammar N2: would to be--> fix grammar N2, N3: So C correct answer--> fix grammar

Thank you for the thorough review and all the suggestions to improve the definition. I have fixed that. Please review it again.

aarunku5:
Thanks, looks good to me @swarooprm

Palipoor (approved):

danyaljj:
@Palipoor @cosmicishan looks like this PR has added two files to the ROOT directory
Screen Shot 2021-10-11 at 12 50 57 AM
. 


@cosmicishan at your earliest convenience, could you send a PR to fix the issue?

PR #390, opened 2021-10-08
Task 1196-1217: ATOMIC

yeganehkordi:
Based on the: https://arxiv.org/pdf/2010.05953.pdf
I've created another task that gives the nodes and asks the relation. But I wasn't sure to add it; I think a classification task with 23 classes might be confusing.


aarunku5:
Thanks! A general comment for your tasks- the definitions need to be rephrased a little to make the task clearer, i.e.-
1) Make it clear that the tuple is split into head and tail (looks like 3 fields in the current phrasing)
2) Can you clarify in the case of personY in the definition, it is essentially any other person mentioned who isn't X?
3) Can you elaborate on the nature of  possible text in the input's head and tail more for tasks? Tasks like reason for, causes, is before, is after, etc. don't need it, but for the rest, the definition doesn't make full sense until you take a look at the examples.

More specific comments:

1197- Specify exactly what you mean by 'feeling' in the definition- I'm understanding that physical feeling is not what is covered in this task based on your examples?

1198- In N1: "input": "Head: PersonX accepts PersonY's offer<sep>Tail: relieved" if Y has made X an offer, Y will be relieved when it is accepted? (I can see a few similarly ambiguous instances as well)

1202- In the instances: "input": "Head: PersonX increases PersonY's rate<sep>Tail: none" --> not clear why this is "Yes"? (there are a few similar instances as well)

1213- In N1: "input": "Head: PersonX affords PersonY protection<sep>Tail: gets involved" --> if x wants to protect y, they would desire to get involved in y's protection so "yes" would be correct? (there are a few similar instances as well)

ashok-arjun:
@yeganehkordi @danyaljj I have the same set of tasks ready, in a similar format. Can I submit it here too?

yeganehkordi:
> @yeganehkordi @danyaljj I have the same set of tasks ready, in a similar format. Can I submit it here too?

As @danyaljj mentioned in the issues, I think you can submit them either. 
Also, I've got another suggestion. With a minor change, we can have some tasks that are given a head node and relation, and asking the tail node. For example, in xIntent relation, the task is to answer the question: "why does PersonX willingly cause the event?"

yeganehkordi:
@aarunku5 Thanks for your comments! I enhanced the definitions and explained the specific comments in the explanations.

danyaljj:

swarooprm:
@aarunku5, check if you are happy or have any other feedback?

aarunku5:
LGTM 👍

PR #371, opened 2021-10-04
Task1186: natural-ness evaluation

Sujan242:
Based on https://arxiv.org/pdf/2109.02938.pdf 
Suggested in issue #286

PR #393, opened 2021-10-09
Tasks 1283-1284: Evaluating quality and Informativeness of System Generated Reference

Sujan242:
Based on https://researchportal.hw.ac.uk/en/datasets/human-ratings-of-natural-language-generation-outputs
The task for naturalness has already been added in #371

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -1119,7 +1119,6 @@ Name | Summary | Category | Domain | Input Language | Output Language
 `task1193_food_course_classification` | Given the name of an indian dish, classify it's course as main course, dessert or snack  | Classification
 `task1194_kth_largest_element` | Given a list of integers and an integer k, find the kth largest element in the list  | Classification
 `task1195_disflqa_disfluent_to_fluent_conversion` | Given a disfluent sentence, modify it to make it a fluent sentence | Text Modification
-
 `task1196_atomic_classification_oeffect` | Given a tuple, determine whether, as a result of the Head, personY or others will be affected as mentioned in the Tail or not. | Classification | Sociology	| English	| English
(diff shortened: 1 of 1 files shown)


Palipoor:
Hi! Thanks for the contribution.
Can you please explain the differences between 2, 3, 4, 5? Or maybe provide examples with the correct answer being 2,3,4,5?

Sujan242:
Thanks for the review! 
I have included an example for each of the ratings from 1 to 6. 
Let me know your thoughts on this.

danyaljj:

danyaljj:
Thanks @Sujan242! 

The issue with this task is that the quality judgments (1-6) are quite subjective and likely noisy. I might label something 3 today and after a few days, I might consider it a 2. 

To amend this issue, I suggest you collapse the labels to, say, 2 labels (`high-score` and `low-score`).

Sujan242:
Thanks for the review @danyaljj. 
The dataset contains 3 judgments given by three different human annotators for each instance. The label that I have added is the median of these 3 scores. Will that not eliminate the noise to a certain extent? 
Also, if I consider the minimum and maximum of these 3 scores, I might be including an outlier. 
Please let me know your thoughts on this

danyaljj:
> Thanks for the review @danyaljj. The dataset contains 3 judgments given by three different human annotators for each instance. The label that I have added is the median of these 3 scores. Will that not eliminate the noise to a certain extent? 

It does to some extent. But there is so much subjectivity about ranking items from 1-6 that the median of 3 judgments might not be enough to suppress the noise. I think think that we should binarize this task. 

Side note: if we can't get this PR in this week, it might not make it since we're catching a deadline soon.

Sujan242:
So should I set a threshold of 3 and everything below that is a 0 and everything above is 1 and update the task definitions accordingly?

danyaljj:
Thanks!

PR #405, opened 2021-10-10
Tasks 1334-1339

swarooprm:
Reviews: swarooprm/CSE576HW#3


Palipoor:
@swarooprm @danyaljj I don't feel good about including tasks 1336 and 1337 in the dataset. I fixed some issues with task 1336 but I don't feel like they're good tasks. What do you think?

Also, in task 1335, I tried fixing the \u.. characters but some of them seem to be meaningless in the context. Like the output of the first positive example is:
`Â¿A quÃ© gÃ©nero se reasignÃ³ el aviÃ³n. `

swarooprm:
> @swarooprm @danyaljj I don't feel good about including tasks 1336 and 1337 in the dataset. I fixed some issues with task 1336 but I don't feel like they're good tasks. What do you think?
> 
> Also, in task 1335, I tried fixing the \u.. characters but some of them seem to be meaningless in the context. Like the output of the first positive example is: `Â¿A quÃ© gÃ©nero se reasignÃ³ el aviÃ³n. `

I agree, I also had similar concerns in the review. Feel free to drop tasks 1336 and 1337 if we can not convert them to good tasks. Let's see what @danyaljj thinks.

danyaljj:
Agreed. Let's drop the tasks if were not happy about their quality. 
At this point, we should optimize quality over quantity.

swarooprm:
I just saw the following description in the huggingface dataset card of the dataset associated with these tasks:
"Automatic machine learning systems can inadvertently accentuate and perpetuate inappropriate human biases. Past work on examining inappropriate biases has largely focused on just individual systems and resources. Further, there is a lack of benchmark datasets for examining inappropriate biases in system predictions. Here, we present the Equity Evaluation Corpus (EEC), which consists of 8,640 English sentences carefully chosen to tease out biases towards certain races and genders. We used the dataset to examine 219 automatic sentiment analysis systems that took part in a recent shared task, SemEval-2018 Task 1 Affect in Tweets. We found that several of the systems showed statistically significant bias; that is, they consistently provide slightly higher sentiment intensity predictions for one race or one gender. We make the EEC freely available, and encourage its use to evaluate biases in sentiment and other NLP tasks."

In ML literature, the popular model debiasing methods require knowledge of dataset biases which is usually found using heuristic and is often difficult to identify (https://arxiv.org/abs/2011.03856), so I was wondering if the tasks in 1336 and 1337 will be useful in that aspect to train a bias-only model?
I don't think we have any other tasks in the repo that cover this aspect
It would be great if we can fix issues we have in these tasks, but if they can't be repaired, let's drop them.

danyaljj:
Classifying a sentence to a race sounds like a problematic thing to learn (dropped it). 
Otherwise, looks good!

PR #407, opened 2021-10-11
Tasks 1344-1349

swarooprm:
Reviews: swarooprm/CSE576HW#12


Palipoor:
So there're a couple of comments That I couldn't fix by simple editing:
1. The QQP examples are not quite good paraphrases. Maybe they haven't been filtered to only include those that are annotated as paraphrases.
2. There're some weird texts in 1348 and 1349 which can be found by searching "entailment" in the files. I don't know what they are.

danyaljj:

danyaljj:
> The QQP examples are not quite good paraphrases. Maybe they haven't been filtered to only include those that are annotated as paraphrases.

Yeah, I have worked with this dataset. Unfortunately, the data is medium-quality. But I tend to be fine with it, given that it has been adopted by many practitioners. 

> There're some weird texts in 1348 and 1349 which can be found by searching "entailment" in the files. I don't know what they are.

You're right, these tasks are not parsed correctly. 
@aritC wondering if you still have your old scripts correct these instances?

aritC:
Hi, @danyaljj I was not aware of this issue up until now. I checked my data and I have the same issue with my scripts as well. 
After googling around for a while I think it may be the case that there are some characters which "pandas" is not able to parse properly and hence it created those weird texts and that it can also be due to Pandas using C engine by default. So, once I changed the engine to python like follows :
`qnli_data = pd.read_csv('../../QNLIv2/QNLI/train.tsv',sep='\t',error_bad_lines=False,index_col=0, engine='python' )`
it is skipping all the erroneous lines correctly and I did not see those lines in my generated JSON anymore.

Hope this helps.

danyaljj:
> Hi, @danyaljj I was not aware of this issue up until now. I checked my data and I have the same issue with my scripts as well. After googling around for a while I think it may be the case that there are some characters which "pandas" is not able to parse properly and hence it created those weird texts and that it can also be due to Pandas using C engine by default. So, once I changed the engine to python like follows : `qnli_data = pd.read_csv('../../QNLIv2/QNLI/train.tsv',sep='\t',error_bad_lines=False,index_col=0, engine='python' )` it is skipping all the erroneous lines correctly and I did not see those lines in my generated JSON anymore.
> 
> Hope this helps.

Got it. Are you able to share the fixed files, say, via a Pull Request to Swaroop's branch?

danyaljj:
Dropped the two problematic tasks (1348 and 1349) and merging the remainder.

PR #426, opened 2021-10-11
Tasks 1403-1406

Ravsehajsinghpuri:


Ravsehajsinghpuri:
@swarooprm there are some conflicts again, can you please look into it, thanks!

swarooprm:
> @swarooprm there are some conflicts again, can you please look into it, thanks!

resolved now. For future references, you can click on "resolve conflict" option and then use the "web editor" option to resolve conflict yourself by deleting few lines that cause conflict.

swarooprm:
@Ravsehajsinghpuri can you fix test failures (repeated example)?

Ravsehajsinghpuri:
> > @swarooprm there are some conflicts again, can you please look into it, thanks!
> 
> resolved now. For future references, you can click on "resolve conflict" option and then use the "web editor" option to resolve conflict yourself by deleting few lines that cause conflict.

Sure! thanks for letting me know about this :)

swarooprm:
Also, I am not sure if task 1404 and 1405 should be different tasks, they are essentially doing the same thing. How about merging these 2 as a single task.
In the definition you can mention, if given format is this, convert to this and vice-versa.
Then in the input, you can also specify format along with the input.

Ravsehajsinghpuri:
> Also, I am not sure if task 1404 and 1405 should be different tasks, they are essentially doing the same thing. How about merging these 2 as a single task.
> In the definition you can mention, if given format is this, convert to this and vice-versa.
> Then in the input, you can also specify format along with the input.

makes sense, instead of merging should I just delete one of them to keep things easy?

swarooprm:
> > Also, I am not sure if task 1404 and 1405 should be different tasks, they are essentially doing the same thing. How about merging these 2 as a single task.
> > In the definition you can mention, if given format is this, convert to this and vice-versa.
> > Then in the input, you can also specify format along with the input.
> 
> makes sense, instead of merging should I just delete one of them to keep things easy?

I would say merge it, it will be a nice task then.
A single unmerged task will be of low quality and contain the risk of rejection in the final dataset.

Ravsehajsinghpuri:
> > > Also, I am not sure if task 1404 and 1405 should be different tasks, they are essentially doing the same thing. How about merging these 2 as a single task.
> > > In the definition you can mention, if given format is this, convert to this and vice-versa.
> > > Then in the input, you can also specify format along with the input.
> > 
> > 
> > makes sense, instead of merging should I just delete one of them to keep things easy?
> 
> I would say merge it, it will be a nice task then. A single unmerged task will be of low quality and contain the risk of rejection in the final dataset.

got it, will do that.

Ravsehajsinghpuri:
> Also, I am not sure if task 1404 and 1405 should be different tasks, they are essentially doing the same thing. How about merging these 2 as a single task.
> In the definition you can mention, if given format is this, convert to this and vice-versa.
> Then in the input, you can also specify format along with the input.

should I be providing the format to which it needs to convert or both the input format and the output format. I think we should provide both in the input so that it can read the input format and then change it to what is required.

swarooprm:
> > Also, I am not sure if task 1404 and 1405 should be different tasks, they are essentially doing the same thing. How about merging these 2 as a single task.
> > In the definition you can mention, if given format is this, convert to this and vice-versa.
> > Then in the input, you can also specify format along with the input.
> 
> should I be providing the format to which it needs to convert or both the input format and the output format. I think we should provide both in the input so that it can read the input format and then change it to what is required.

I think you can just provide the input format, you don't have to provide the output format. Mention in the definition that the task is to flip format i.e if it is in MM/DD, convert it to DD/MM and vice-versa. @Ravsehajsinghpuri

Ravsehajsinghpuri:
> > > Also, I am not sure if task 1404 and 1405 should be different tasks, they are essentially doing the same thing. How about merging these 2 as a single task.
> > > In the definition you can mention, if given format is this, convert to this and vice-versa.
> > > Then in the input, you can also specify format along with the input.
> > 
> > 
> > should I be providing the format to which it needs to convert or both the input format and the output format. I think we should provide both in the input so that it can read the input format and then change it to what is required.
> 
> I think you can just provide the input format, you don't have to provide the output format. Mention in the definition that the task is to flip format i.e if it is in MM/DD, convert it to DD/MM and vice-versa. @Ravsehajsinghpuri

yes, I have done exactly the same thing and pushed the recent commit. Please review and lmk

tasks/task1403_check_validity_date_mmddyyyy.json
```diff
@@ -0,0 +1,1217 @@
...
+    ], 
+    "Definition": "In this task, you are given a date in \"mm/dd/yyyy\" format. You need to check if the date is valid or not. Return 1 if it is valid, else return 0. A date is valid is the components month(\"mm\"), day(\"dd\") and year(\"yyyy\") are all valid individually. A day(dd) is valid if it is greater than or equal to 1 and less than 30 or 31 depending upon the month(mm). Months which have 31 days are January, March, May, July, August, October, December. Rest of the months have 30 days except February which has 28 days if it is not a leap year and 29 days if it is a leap year. A month(mm) is valid if it lies in the range from 1 to 12 as there are 12 months in a year. A year is always valid if it is expressed in the form of \"yyyy\".", 
+    "Positive Examples": [
+        {
+            "input": "14/25/1405", 
+            "output": "0",
+            "explanation": "It is an invalid date as the month(mm) is 14 which does not lie in the range 1 to 12."
+        }, 
+        {
+            "input": "7/29/1617",
```
Palipoor: Here, for (mm) or (dd) if they're single-digit numbers, shouldn't they appear as, say, 07 instead of 7?

Ravsehajsinghpuri: yes, you're right - I have made the changes, pls review

Palipoor (approved):

Issue #383, opened 2021-10-07
Classification [im]balance 


danyaljj:
In the test file, automatically detect classification tasks and flag those that have too large of skew.

PR #427, opened 2021-10-11
Update test_all

yeganehkordi:
Update tests based on #383 
@danyaljj  I'm not sure about the threshold; I thought over 0.5 would be too large.


.github/workflows/blank.yml
```diff
@@ -31,6 +31,10 @@ jobs:
         with:
           path: "src/requirements.txt"
 
+      # Install Dependencies
+      - name: Install Dependencies
+        run: sudo pip install scipy
```
danyaljj: Add the dependency to `requirements.txt`

yeganehkordi: Sure!

src/test_all.py
```diff
@@ -127,7 +128,19 @@ def assert_language_name(name):
                     except KeyError:
                         raise Exception(f" * Looks like we have a repeated example here! "
                                         f"Merge outputs before removing duplicates. :-/ \n {instance}")
-
+                        
+                        
+            #Make sure classes are balanced
+            if 'Classification' in data['Categories']:
```
danyaljj: I think it's better to infer this automatically (rather than checking for the classification tag in the categories)
Basically, find the set of unique category labels: 
```
set( [x['output'] for x in data['Instances'][:100] )`
```
 and check it is small (say, less than 15 categories).

src/test_all.py
```diff
@@ -127,7 +128,19 @@ def assert_language_name(name):
...
+                outputs=[ins['output'] for ins in instances]
+                classes = []
+                for i in outputs:
+                    if i not in classes:
+                        classes.append(i)
+                output_int=[classes.index(output) + 1 for output in outputs]
+                assert abs(skew(output_int)) < 0.5, "Classes distribution is skewed"
```
danyaljj: I am actually not sure if the `skew` function is the right choice here. 

>  For **unimodal continuous distributions**, a skewness value greater than zero means that there is more weight in the right tail of the distribution.

yeganehkordi: I guess that would be fine. Since we might have negative values as well, I used abs().

danyaljj: Consider a simpler metric, like 
- relative popularity of a particular category compared to an average 
- a measure based on entropy might also work (since entropy measures some notion of similarity to uniform distributions)

yeganehkordi: Sure! I'll fix it tomorrow.

src/test_all.py
yeganehkordi: > rather than checking for the classification tag in the categories

I'm not sure if I get it correctly; you mean to run this script on all the tasks?

danyaljj: Ya

danyaljj: If it is not too slow, I think it's worth the shot.

src/test_all.py
yeganehkordi: Oops! I just get your point. yeah, this function is not working correctly.

src/test_all.py
yeganehkordi: Thanks! Let's try the speed.

src/test_all.py
```diff
@@ -33,6 +35,18 @@
...
+    value,counts = np.unique(result, return_counts=True)
+    norm_counts = counts / counts.sum()
+    entropy=-(norm_counts * np.log(norm_counts)/np.log(len(value))).sum()
+    assert entropy > 0.7, "Classes distribution is skewed"
+
+def skewness2(result):
+    value,counts = np.unique(result, return_counts=True)
+    average=np.average(counts)
+    metric=np.average(abs(counts-average)/counts.sum())
+    assert metric < 0.2, "Classes distribution is skewed"
```
danyaljj: For now, suggest just printing the skew, rather than failing. We'll need to pick a threshold based on the outcome.

yeganehkordi: I wasn't sure which metric would be better. Should I print the skew based on the both metrics?

danyaljj: Yes! print them side by side. Also print the distribution as well.

yeganehkordi: Sure! I'll fix it now.

danyaljj:
How can the following be the distribution? 
If the task is a binary classification, we should have an array of size 2; right? 
```
 --> testing file: task007_mctaco_answer_generation_transient_stationary.json
⚠️ WARNING: did not find the key: Domains
Distribution of classes: [  1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1
   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1
   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1
   1   1   1   2   1   1   1   1   1   1 115   1   1   1   1   1   1   1
   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1
   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1
   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1 135
   1   1]
```

src/test_all.py
```diff
@@ -140,7 +157,14 @@ def assert_language_name(name):
                     except KeyError:
                         raise Exception(f" * Looks like we have a repeated example here! "
                                         f"Merge outputs before removing duplicates. :-/ \n {instance}")
-
+                        
+                        
+            #Make sure classes are balanced
+            output=[ins['output'] for ins in instances]
+            outputs = sum(output, [])
```
danyaljj, line 202: What is the purpose of this line?

src/test_all.py
```diff
@@ -33,6 +35,21 @@
 def assert_language_name(name):
     assert name in language_names, f"Did not find `{name}` among iso639 language names: {language_names}"
 
+def skewness(result):
+    value,counts = np.unique(result, return_counts=True)
+    norm_counts = counts / counts.sum()
+    entropy=-(norm_counts * np.log(norm_counts)/np.log(len(value))).sum()
+    if entropy < 0.7:
+        print(f'Distribution of classes: {counts}')
```
danyaljj: print `norm_counts`

yeganehkordi:
> If the task is a binary classification, we should have an array of size 2; right?


No, it's an answer generation task. As we have lots of "yes" and "no" labels in the answers, it seems to be skewed based on the entropy metrics.

src/test_all.py
yeganehkordi: Each `ins['output']` is a list, so I get the real outputs in the following line.

src/test_all.py
yeganehkordi: Just now!

src/test_all.py
danyaljj: Got it. Maybe add a comment just above the code that indicates that this line "flattens the nested arrays"

src/test_all.py
```diff
@@ -140,7 +158,14 @@ def assert_language_name(name):
                     except KeyError:
                         raise Exception(f" * Looks like we have a repeated example here! "
                                         f"Merge outputs before removing duplicates. :-/ \n {instance}")
-
+                        
+                        
+            #Make sure classes are balanced
+            output=[ins['output'] for ins in instances]
+            outputs = sum(output, [])
+            skewness(outputs)
```
danyaljj: Before running these skewness tests, you'd need to check if the number of unique outputs is not too many (say, it's less than 15).

src/test_all.py
```diff
@@ -33,6 +35,22 @@
...
+    if entropy < 0.7:
+        print(f'Norm_counts: {norm_counts}')        
+        print(f'Distribution of classes: {counts}')
+        print(f'⚠️ WARNING: Classes distribution is skewed, entropy= {entropy}.')
+
+def skewness2(result):
+    value,counts = np.unique(result, return_counts=True)
+    average=np.average(counts)
+    metric=np.average(abs(counts-average)/counts.sum())
+    if metric > 0.2:
```
danyaljj: drop these conditions for now and print the two skewness measures side by side.

src/test_all.py
```diff
@@ -33,6 +35,22 @@
 def assert_language_name(name):
     assert name in language_names, f"Did not find `{name}` among iso639 language names: {language_names}"
 
+def skewness(result):
+    value,counts = np.unique(result, return_counts=True)
+    norm_counts = counts / counts.sum()
+    entropy=-(norm_counts * np.log(norm_counts)/np.log(len(value))).sum()
```
danyaljj: I understand that what `norm_counts * np.log(norm_counts)` is the equivalent of the following:  
image



But not sure why you have this:  `/np.log(len(value))`

src/test_all.py
yeganehkordi: Sure! Will do.

src/test_all.py
yeganehkordi: I've got a problem here. We had some classification tasks with more than 70 classes, for example task 481 here: https://github.com/swarooprm/CSE576HW/pull/111
Although checking some of the classes gives us a sense of skewness, but it might be inaccurate. On the other hand, checking all the different outputs in other categories is not necessary.

src/test_all.py
yeganehkordi: I will.

src/test_all.py
yeganehkordi: If we choose a single base for all the tasks, we can't pick a threshold because the upper bound will increase with the number of classes. That's why I used the number of classes as a base here.

danyaljj: I see. That makes sense!

src/test_all.py
danyaljj: How about we try a mix of the two approaches you have tried: 
```
if the number of unique outputs < 15 or 'classification' is among the  categories
    ... 
```

yeganehkordi: Yeah, this is the best way. I'll change it now.

danyaljj:
@yeganehkordi could you create a Google spreadsheet of: tasks, their skew measures and their norm-counts? (basically, same content printed out in the logs, but in a spreadsheet so that we can sort and compare them).

yeganehkordi:
Yes, I'll create it now. I was writing some tasks that are not correlated with our thresholds in the comments. Here are some examples:
- Task 043: [0.08537194 0.09985022 0.10034948 0.09835247 0.61607589], entropy= 0.74, metric= 0.16 
- Task 56: [0.27392739 0.72607261], entropy= 0.84, metric= 0.22
- Task 1154: [185  31  62  62 125 217  31  31  31  31  30  94  64], entropy= 0.89, metric= 0.04
- Task 147: [0.8085 0.1915], entropy= 0.70, metric= 0.30
- Task155: [ 680 1917    3    1    1 1018 1262 1003  435  125   38   10    7], entropy= 0.70, metric= 0.08

danyaljj:
@yeganehkordi don't copy the numbers one by one. Print everything in a tsv file (each line`{file_name} \t {skew1} \t {skew2} \t {dist} \n`) and just copy-paste the whole content to the Google sheet.

yeganehkordi:
Nice :) Thanks for saving me!

danyaljj:
Summary thus far: 
 - @yeganehkordi has created two metrics that are quite correlated. So we can safely use one of them (say, the one based on entropy).  
 - Based on these metrics, she has identified 20 and something tasks that are heavily skewed. https://docs.google.com/spreadsheets/d/1AYWKvyF9ufSA-zkl6Fv951a2Q48VfdHSv-n8pJl5Dl8/edit?usp=sharing
We need to address these, before we can any further progress on this PR.

yeganehkordi:
@danyaljj I can address them.

danyaljj:
When we get back to this PR next week or so: 
Raise flag if the size of the unique outputs is 1; it must be a bug (appeared in task201 #468 )

yeganehkordi:
Sure!

danyaljj:
To move on with this PR, I have created an exclusion list to skip over the old skewed tasks. 
Over time, we should fix these skew tasks and drop the exclusion list.

yeganehkordi:
> When we get back to this PR next week or so: Raise flag if the size of the unique outputs is 1; it must be a bug (appeared in task201 #468 )

@danyaljj Sorry, I didn't address this comment. I thought we will merge this PR after merging skewed tasks. I'll fix it.

danyaljj:
Oh already did: https://github.com/allenai/natural-instructions-expansion/blob/master/src/test_all.py#L214

yeganehkordi:
Thanks and sorry again!

PR #433, opened 2021-10-12
Task 1418: bless lexical semantics task

atharva-naik:
Added task1418: classifying lexical relation between given pair of words. The instances are created out of the BLESS dataset, which was first proposed in [this](https://aclanthology.org/W11-2501.pdf) paper.


swarooprm:
Fixed source=list error, add the following fields to your file to resolve the current error
"Input_language": [
"English"
],
"Output_language": [
"English"
],
"Instruction_language": [
"English"
]

swarooprm:
This task is good to go once labels in the instances are balanced.

Issue #384, opened 2021-10-07
Decomposing math problems into steps 


danyaljj:
Down the line when we support structured outputs: 
https://math-qa.github.io/math-QA/

For now, we can build simpler tasks based on this.

mishra-sid:
I am making simple classification tasks based on this.

PR #438, opened 2021-10-13
Task 1419 - 1424 : MathQA

mishra-sid:
MathQA: [MathQA](https://math-qa.github.io/) 
Task originally suggested in #384 
Tasks derived from MathQA Datasets for each of the categories of problems.

Explanations are derived from the "Rationale" field annotated by MathQA, hence they need not be "mathematically" verified.


mishra-sid:
@Palipoor Thanks for the review and changes made to the task definition, looks much better now. Let me know if there are any other changes that I need to make before It can be merged.

danyaljj:

PR #443, opened 2021-10-13
Add sub-categories for verification

Palipoor:
I added these sub-categories as I have seen related tasks:
```
  - `Verification -> Summary Verification`
  - `Verification -> Subject Verification`
 ```

### DIFF
--- a/doc/task-hierarchy.md
+++ b/doc/task-hierarchy.md
@@ -84,6 +84,8 @@
 - `Translation`
 - `Verification`: Verify if two inputs match with each other e.g. a task to verify if a sentence and its sentiment given in the input match
   - `Verification -> Claim Verification`
+  - `Verification -> Summary Verification`
+  - `Verification -> Subject Verification`
 - `Weblink Generation`
 - `Word Sense Disambiguation`
 


danyaljj:
Actually, a basic question: what do mean by "verification"? 
And building on that answer, what are the "summary" and "subject" verification variants? 

FYI @ghlai9665  @swarooprm

Palipoor:
I think it refers to the tasks where a pair is given, and the answer needs to verify whether a certain relationship between them exists. Like: Is the second question a correct paraphrase of the first one? Or: Is the given summary a good summary for the given text?
It can be categorized as "classification" too, but I think verification might be better.

garyhlai:
@Palipoor perhaps you could link us to all the tasks where you think these categories apply so we could understand better?

@danyaljj How do you plan to use these categories in the later experiments? I think it's important to think about how these categories are gonna be used to determine whether they are meaningful or not in the task hierarchy.

danyaljj:
> @Palipoor perhaps you could link us to all the tasks where you think these categories apply so we could understand better?

Let me try to take a stab at this: 

To me, verification is a special version of classification (*) when one needs to **verify** whether a given descriptive attribute (**) applies to a given text or not (binary output). 

I distinguish verification from binary (yes/no) QA because: 
 - in verification, the attribute being verified is defined globally across all the instances 
 - in binary (yes/no) QA, the attribute being question is specific for each instance (each instance comes with a different question). 


(*) if that's the case, we'd need to have `Classification -> Verification` 
(**) such as checking whether a sentence is grammatical or not, checking if the paragraph contains offensive content or not, checking if sentence a is concluded from another sentence or not. 


> @danyaljj How do you plan to use these categories in the later experiments? I think it's important to think about how these categories are gonna be used to determine whether they are meaningful or not in the task hierarchy.

This is a good question. We will use these categories to measure "generalization" across distinct categories. Say, if we use **non-**QA  tasks to train a model, can it solve, say, QA tasks? You can see examples of such experiments in Table 4 of [this paper](https://arxiv.org/pdf/2104.08773.pdf).

Palipoor:
@ghlai9665 
These are some of the tasks that I think fit into "verification".

- task066_timetravel_binary_consistency_classification
- task133_winowhy_reason_plausibility_detection
- task204_mnli_same_genre_classification
- task209_stancedetection_classification
- task211_logic2text_classification
- task227_clariq_classification

garyhlai:
> This is a good question. We will use these categories to measure "generalization" across distinct categories. Say, if we use **non-**QA tasks to train a model, can it solve, say, QA tasks? You can see examples of such experiments in Table 4 of [this paper](https://arxiv.org/pdf/2104.08773.pdf).

@danyaljj Oh! of course 😅 , just read the paper again. I think I'm just not sure how we will use the "task hierarchy" since there was no "task hierarchy" in the paper, just plain categories.

danyaljj:
Closing it in favor of #459

PR #449, opened 2021-10-15
Tasks 1443 - 1446: Synthetic data

kurbster:


kurbster:
@yeganehkordi What is the list of domains? For tasks 1444-1446 I put the domain as mathematics as those tasks involve rounding, and subtraction. But what should I put for task 1443? It does not really involve mathematics simply parsing strings into numbers.

I have made the change by removing the comma and I will push that once I add the domain to 1443. Thank you!

kurbster:
I found the list of Domains, I am still unsure what to use for task 1443. Maybe mathematics?

yeganehkordi:
Thanks! Yeah, I think mathematics is the best match.

swarooprm (approved):

Issue #217, opened 2021-09-09
ATOMIC


danyaljj:
One can create tasks based on the atomic data, where the task definitions define the edges/relations in their dataset:
https://allenai.org/data/atomic-2020

swarooprm:
This one is huge and has a lot of varieties, many tasks can be done using this. Same is true for Conceptnet.

yeganehkordi:
I looked at this dataset. I thought of defining the edges/relations task and several classification tasks based on the meaning (Physical-Entity Commonsense, Social-Interaction Commonsense, and Event-Centered Commonsense) and the relations. I might be able to define 27 tasks based on this dataset. How many tasks and which tasks are desirable?

swarooprm:
You are right! Different relations produce different tasks.
Re: How many tasks: No such estimate, more the better :)

ashok-arjun:
@yeganehkordi If you're still working on this and need help, I can collaborate and take up some tasks

yeganehkordi:
@ashok-arjun Thanks! I haven't started working on the classification tasks based on the meaning (Physical-Entity Commonsense, Social-Interaction Commonsense, and Event-Centered Commonsense). Feel free to take them.

ashok-arjun:
Can you explain what you mean by classification tasks? @yeganehkordi Because all I see in the dataset are such tasks? 

Does this also mean you've also worked on some tasks from the dataset?

ashok-arjun:
@swarooprm @danyaljj Would physical entity tasks be defined in the following way be acceptable? I'm not sure since the input isn't a sentence

**Task**: Predict what the entity is made up of 

**Input**: Bread
**Output**: Dough, wheat

danyaljj:
> @swarooprm @danyaljj Would physical entity tasks be defined in the following way be acceptable? I'm not sure since the input isn't a sentence
> 
> **Task**: Predict what the entity is made up of
> 
> **Input**: Bread **Output**: Dough, wheat

Yeah I think that should be good!

danyaljj:
Just have this in mind that: small high-quality data is better than large low-quality data.

ashok-arjun:
Sure - I'll hand-pick about 15-20 required tasks and work on them.

yeganehkordi:
> Can you explain what you mean by classification tasks? @yeganehkordi Because all I see in the dataset are such tasks?
> 
> Does this also mean you've also worked on some tasks from the dataset?

[...]

PR #453, opened 2021-10-16
Task 1454 - 1478: Conceptnet English

mishra-sid:
Tasks derived from edges of Conceptnet , A common sense knowledge graph ( https://conceptnet.io/ )
Similar to ATOMIC tasks in #217 and the corresponding PR in #390  , but a larger data set with more relations and different format/data source. They can be reviewed/merged together.


mishra-sid:
@danyaljj @swarooprm I have contributed a total of 34 tasks (including this PR, in  PoS tagging, Question Answering and Knowledge graphs). Although, my total number of commits (after merge) is only 5 since all the PRs were/would be rebased and hence I'm not in one of the top 20 contributors commits wise. Would It still count as a "Significant Contribution" ?  
 I have reviewed a couple PRs yet, but am happy to help with the reviewing of the remaining open PRs if assigned and can also help with cleaning up of tasks.

swarooprm:
@kuntalkumarpal, can you review these tasks?

swarooprm:
@Mehrad0711, would you have time to review this PR?

Mehrad0711:
Some of the terms used in task instructions need explanation. Instructions need to be clear, complete, and concise. If the task is confusing for a human it would definitely be for the neural model.
Some of the provided examples don't quite make sense. Please see the details below. I think some filtering is needed to improve the quality of these datasets.
Given "First term:" and "Second term:" are constant across all examples I suggest removing them. It'll speed up training/ inference without accuracy penalty.

`task1454_conceptnet_form_of`:
Expand what "inflected form" of a term means. 

`task1455_conceptnet_derived_from`:
Explain exactly what "derived" means. Is it physical extraction, lexical derivation, a class property, etc. ? Examples suggest all the above.


`task1456_conceptnet_has_context`:
Again what does "in the context" entails here. "o" in "jocular" and "this" in "internet slang" are different categories. Also, some of the examples don't make sense to me. How is "dadt" typically used in "politics"? 

`task1463_conceptnet_has_prerequisite`:
Task definition is too broad. In several examples, the first term does not need to necessarily happen for the second term to happen.
Someone may need a haircut just because they prefer to keep it short and clean. It doesn't have to be too long. Furthermore, examples like:
```
"input": "First term: saxophone \nSecond term: saxophone",
"output": [
    "No"
]
```
are confusing. If I really need to answer, I would lean towards a "Yes" than a "No". But this is an existential question rather than a temporal cause-effect.


I suggest addressing the above before I can review the remaining tasks. Specifically, all task definitions need to be updated. Keywords (such as "inflected", "subverted") need a clear explanation. Examples need some cleanup - this is tricky and I'm not sure how besides manual reviewing. Some heuristics may be used to remove ambiguous examples.

mishra-sid:
Thanks for the feedback.

Regarding the first point: I think this is the pattern generally used across all tasks having more than one input ( Some label for each and a separator between them ), I'll be happy to remove it if needed.  

I do agree with all of the rest of these points. I will work on giving better instructions/task definitions and to get better examples and relevant explanations for each task. Will cut down on ambiguous tasks and instances based on some heuristic that I think is more reasonable to address your comment regarding quality of these datasets.

Mehrad0711:
> Regarding the first point: I think this is the pattern generally used across all tasks having more than one input ( Some label for each and a separator between them ), I'll be happy to remove it if needed.

It's a matter of preference really and how the model will be trained. I think as long as you have the separator, the labels are unnecessary.

aarunku5:
> > Regarding the first point: I think this is the pattern generally used across all tasks having more than one input ( Some label for each and a separator between them ), I'll be happy to remove it if needed.
> 
> It's a matter of preference really and how the model will be trained. I think as long as you have the separator, the labels are unnecessary.

@mishra-sid any updates on your end so we can proceed with this PR?

mishra-sid:
@aarunku5 I'm almost done cleaning up the data and addressing the issues mentioned by Mehrad, should have a PR ready for review by tomorrow.

swarooprm:
> @aarunku5 I'm almost done cleaning up the data and addressing the issues mentioned by Mehrad, should have a PR ready for review by tomorrow.

@mishra-sid, a gentle reminder!

PR #447, opened 2021-10-14
Address human feedback 

Palipoor:
Hi. I will address human feedbacks from #276 here.

PR #455, opened 2021-10-16
Update zest tasks

yeganehkordi:


Palipoor:
Hi!
Do you think we can do it the other way and capitalize the keywords? Because when "DOG BREED" is a keyword and a placeholder for a breed of dog, the sentences make sense, but now they're somehow weird.

danyaljj:
> Hi! Do you think we can do it the other way and capitalize the keywords? Because when "DOG BREED" is a keyword and a placeholder for a breed of dog, the sentences make sense, but now they're somehow weird.

+1

Palipoor:
Actually, after looking closer, I realized in all sentences there're words "this" before the words that were supposed to be capitalized, which makes them definite. For example:
"How did the stock market perform while this president was in office?"
instead of 
"How did the stock market perform while PRESIDENT was in office?"

So I guess, for the sake of simplicity, we can change the examples to be like the first sentence(i.e. lower case + "this")
What do you think? @yeganehkordi @danyaljj

danyaljj:
> Actually, after looking closer, I realized in all sentences there're words "this" before the words that were supposed to be capitalized, which makes them definite. For example: "How did the stock market perform while this president was in office?" instead of "How did the stock market perform while PRESIDENT was in office?"
> 
> So I guess, for the sake of simplicity, we can change the examples to be like the first sentence(i.e. lower case + "this") What do you think? @yeganehkordi @danyaljj


Or: 
```
 - if the keyword already has "the", just .lowercase 
 - otherwise, "the" + .lowercase or "this" + .lowercase
```

yeganehkordi:
Sorry for my delay! We have three keywords in these tasks (president, dog breed, and national park), and all of them have the word "this" in the instances. It seems that they don't come with "the" in the instances. So I think we can explicitly mention the keywords and their format in the definition.
I'll update them now.

danyaljj:
Looks good to me. Leaving the merge call to @Palipoor

tasks/task119_zest_text_modification.json
```diff
@@ -6,46 +6,51 @@
...
+            "input": "Does this dog breed get taller than 2 feet?",
+            "output": "Is this dog breed tall?",
             "explanation": "This question is vague because it didn't specify tallness. Make your questions specific and concrete rather than open-ended or subjective."
         },
         {
-            "input": "Is DOG BREED bigger than 12 inches?",
-            "output": "Which DOG BREED is the biggest?",
-            "explanation": "Do not write questions that compare domains"
+            "input": "Is this dog breed bigger than 12 inches?",
+            "output": "Which this dog breed is the biggest?",
```
Palipoor, line 42: I think this question should be: Which dog breed is the biggest?

Palipoor:
I approved this and will merge it. It will probably cause merging issues when we will merge #447, but it will be ok with accepting the master.

danyaljj:
Yeah, it will cause merge conflicts that require manual intervention. 
In case it helps, (and at the risk of a bit deviation from the topic of this thread) PyCharm makes it pretty easy to resolve merge conflicts: https://www.youtube.com/watch?v=IZW9XJQrwaU

Palipoor:
Thanks! I just accepted Yeganeh's version.

PR #473, opened 2021-10-23
A script to fix unicode characters in files

Palipoor:
A lot of task files contain characters in the \u.. format. I wrote a script to fix them.


danyaljj:
Thanks! When the repository because relatively stable (next or next-next week), we should run this script on all of our tasks.

PR #486, opened 2021-10-26
Tasks 1511-1513

sumantapatro:
Task 1511

### DIFF
--- a/tasks/task1512_country_population.json
+++ b/tasks/task1512_country_population.json
@@ -9,7 +9,7 @@
       "Answer Generation",
       "Factual Question Answering"
    ],
-   "Definition":"In this task, you are given a country name and you need to return the population of the given country, rounded to the nearest 100,000 in the year 2000",
+   "Definition":"In this task, you are given a country name and you need to return the population of the given country, rounded to the nearest 100,000, as of the year 2000",
    "Input_language":[
       "English"
    ],


swarooprm:
@sumantapatro, this PR is showing that 13 files are being changed, I guess you want to push 1 file since the rest of the 12 files are already merged.
Could you fetch upstream and sync before pushing changes?
You can also create a branch instead of merging from the master, this is optional though.

aarunku5:
Thanks!

For tasks 1511, 1512- these aren't exactly human-achievable tasks. Can you modify both tasks to give rounded estimates? For instance- life expectancy can be 60-70 years and population can be 1.2 million

In 1513, define what landlocked is in the definition

aarunku5:
1513 looks good. To clarify, for 1511 and 1512, can you make updates to the examples and instances as follows:

1511
"input":"Angola",
"output":"30-40"
"explanation":"Life expectancy of the given country Angola is between 30-40 in the year 2014"

1512
"input":"Angola",
"output":"12.8 million",
"explanation":"Population of the given country Angola is 12.8 million in the year 2000"

aarunku5:
@sumantapatro any updates on your end so we can proceed with this PR?

sumantapatro:
> @sumantapatro any updates on your end so we can proceed with this PR?

@aarunku5 Task1511 and 1512 are created based on the data-set available in the mentioned URLs as defined in the respective tasks.

Issue #482, opened 2021-10-25
Missing task numbers


swarooprm:
Let's update all the missing task numbers here, we can fill these numbers from the tasks in the latest PRs.
task063-064: https://github.com/allenai/natural-instructions-expansion/pull/474

andersonjwan:
@swarooprm I created a small script to find which tasks are missing from the list. Unfortunately, it is a lot - 314 total. You can run the script below if you are curious yourself.

p.s. The script should be ran in the `tasks/` directory. I've also attached the list of missing task numbers instead of posting a comment to reduce spam [here](https://github.com/allenai/natural-instructions-expansion/files/7428008/missing-tasks.txt).

```python
if __name__ == "__main__":
    with open("README.md", "r", encoding="utf-8") as infile:
        comp = []
        line = infile.readline()

        while line:
            if line[0] == '`':
                task = (line.split("|")[0]).split("_")[0]
                number = task[5:] # get task number

                comp.append(int(number))

                if len(comp) == 2:
                    # do comparison
                    if (comp[1] - comp[0]) > 1:
                        for i in range((comp[1] - comp[0]) - 1):
                            print(f"Missing Task {comp[0] + i + 1}")

                    comp.pop(0)

            line = infile.readline()
```

danyaljj:
Thanks @andersonjwan!

PR #502, opened 2021-10-30
Task 1551 + Missing Task Numbers from Issue #482

pulkitverma25:


aarunku5:
Thanks! General comment- make sure you have at least 2 each of positive and negative examples for all your tasks.
Here are the task-wise comments for definition edits:

63:
Change Definition: "In this task, you are given inputs i and A, where i is an integer and A is a list. You need to output the first i elements of A, starting with the 1st element in A. i will always have a value less than the length of A."

64:
Change Definition: "In this task, you are given inputs i and A, where i is an integer and A is a list. You need to list all the elements of A after the first i elements. i will always have a value less than the length of A."

78:
Change Definition: "In this task, you are given inputs i and A, where i is an integer and A is a list. You need to list all the elements of A preceding the last i elements. i will always have a value less than the length of A."

91:
Change Definition: "In this task, you are given inputs i,j, and A, where i and j are integers and A is a list. You need to list all elements of A from the ith element to the jth element. i and j will be non-negative, and will always have a value less than the length of A. i will always be less than j."

99:
Change Definition: "In this task, you are given inputs i,j, and A, where i and j are integers and A is a list. You need to list all elements of A from the ith element to the jth element in the reverse order. i and j will be non-negative, and will always have a value less than the length of A. i will always be less than j."

100:
Change Definition: "In this task, you are given inputs i,j, and A, where i and j are integers and A is a list. You need to concatenate all elements of A from the ith element to the jth element, and print the resultant string. i and j will be non-negative, and will always have a value less than the length of A. i will always be less than j."

101:
Change Definition: "In this task, you are given inputs i,j, and A, where i and j are integers and A is a list. You need to find all elements of A from the ith element to the jth element, reverse the order in which they occur, and then concatenate them to output the resultant string. i and j will be non-negative, and will always have a value less than the length of A. i will always be less than j. Perform the operations in order, i.e., do not concatenate the elements and then reverse the string."

1551:
Change Definition: "In this task, you are given inputs k,i, and A, where k and i are integers and A is a list. You need to find every ith element of A starting from the kth element. The process stops when the position of the next ith element exceeds the length of the list."

267:
Change Definition: "In this task, you are given inputs i,j, and A, where i and j are integers and A is a list. You need to concatenate all elements of A from the ith element to the jth element, and then reverse the resulting string. i and j will be non-negative, and will always have a value less than the length of A. i will always be less than j. Perform the operations in order, i.e., do not reverse first and then concatenate."

506, 507:
In definitions, edit this: "Return -1 ... **is** in the list"

606:
Change Definition: "In this task, you are given inputs i,j, and A, where i and j are integers and A is a list. You need to return the sum of all the numerical elements in the list A between the positions i and j (including positions i and j). Return 0 if no numerical element is present in the list between the given ranges.  i and j will be non-negative, and will always have a value less than the length of A. i will always be less than j. "

636:
Change Definition: "In this task, you are given an input list A comprising of numbers and alphabets. You need to extract and sort the unique alphabets in the list. The alphabets in the input list will only be in lowercase."

755:
Change Definition: "In this task, you are given two strings A,B. You must perform the following operations to generate the required output list: (i) Find the longest common substring in the strings A and B, (ii) Convert this substring to all lowercase and sort it alphabetically, (iii) Replace the substring at its respective positions in the two lists with the updated substring."

756:
Change Definition: "In this task, you are given two strings A,B. Find the longer of the two lists, convert it to lowercase, and return all the unique alphabets used in it. The two input strings are never equal."

pulkitverma25:
Thanks for the suggestions @aarunku5. I have updated the definitions as suggested and provided at least 2 examples each for both positive and negative cases for all tasks.

aarunku5:
LGTM 👍

aarunku5 (approved):

PR #504, opened 2021-10-30
Tasks 1547-1550

swarooprm:
Reviews: https://github.com/swarooprm/CSE576HW/pull/73


aarunku5:
@swarooprm 1547- The question format is a little odd in the input, can you cross-check it?

PR #553, opened 2021-11-02
Tasks 1737-1744

swarooprm:
Reviews:
https://github.com/swarooprm/CSE576HW/pull/118
https://github.com/swarooprm/CSE576HW/pull/110


swarooprm:
1743: examples should not be a list.

Issue #522, opened 2021-11-01
XCOPA bug 


danyaljj:
@amirrezamirzaei  In `task1179_xcopa_commonsense_cause_effect_th`, everything is mapped to `effect`.

amirrezamirzaei:
yes, I looked at the data it seems that they have been labeled incorrectly in the source [dataset](https://github.com/cambridgeltl/xcopa). I will try to label them correctly myself.

danyaljj:
I see. If it's something that will take more than 30 mins, don't waste your time on it (It might be easier to drop/replace the task.)

amirrezamirzaei:
I think we can close this issue.

PR #564, opened 2021-11-05
fix xcopa_cause_effect_th

amirrezamirzaei:
addressing #522

### DIFF
--- a/tasks/task1171_xcopa_commonsense_cause_effect_id.json
+++ b/tasks/task1171_xcopa_commonsense_cause_effect_id.json
@@ -535,7 +535,7 @@
         {
             "input": "Tinta di poster itu luntur.\nSaya mencipratkan air ke poster.",
             "output": [
-                "effect"
+                "cause"
             ]
         },
         {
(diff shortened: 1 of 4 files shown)


danyaljj:
@amirrezamirzaei could you briefly explain the fix?

amirrezamirzaei:
of course, the dataset uses the same sentences for each language so I just used another language to find the labels.

amirrezamirzaei:
something else I noticed was that the output for the reasoning tasks is exactly like each other for every language. 
but for the cause and effect task, they vary a little (around 20 to 60) between some languages should I evaluate these instances by hand?

danyaljj:
> something else I noticed was that the output for the reasoning tasks is exactly like each other for every language. but for the cause and effect task, they vary a little (around 20 to 60) between some languages should I evaluate these instances by hand?

Well, somehow we need to verify the quality. Unfortunately, none of our contributors speak Thai. I wonder if we can use Google translate to verify a random subset of, say 20 instances?

amirrezamirzaei:
most of the outputs were like each other so I used the first task (969) to check and fix all the differences with the help of google translate.

danyaljj:
Thanks!

PR #586, opened 2021-11-10
added language tags for tasks 938-954

### LINKED ISSUE #572: Missing language tags
are missing for the following tasks: 
 - 938-943 
 - 944-954 

@tanay2001 could you address this issue?

tanay2001:
fixes #572


tanay2001:
> fixes #572

not sure if the tests are failing cuz of the changes I have made, should I drop the task which is failing or the language name has to be something else ? would be great if you could help 
@danyaljj

danyaljj:
Based on the fail log, `Odia` is not among ISO languages. But it looks like `Oriya` is an equivalent name that we have.

danyaljj:
Punjabi -> Panjabi

tanay2001:
> Punjabi -> Panjabi

yup , thanks a lot fixed it !

PR #595, opened 2021-11-11
Update languages for tasks 927 - 1088 and address some human feedback.

Palipoor:
Some of the human feedbacks* didn't make sense. My guess is that it had something to do with the unfamiliarity of the languages of the tasks, and/or the characters not rendering correctly.  

* 938 - 952

### DIFF
--- a/tasks/task927_yelp_negative_to_positive_style_transfer.json
+++ b/tasks/task927_yelp_negative_to_positive_style_transfer.json
@@ -8,49 +8,49 @@
     "Categories": [
         "Sentence Generation"
     ],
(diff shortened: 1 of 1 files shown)


danyaljj:
Thanks!  
Yeah, there were a few mistakes on my side that caused some issues in rendering the examples, unfortunately.

tanay2001:
> Some of the human feedbacks* didn't make sense. My guess is that it had something to do with the unfamiliarity of the languages of the tasks, and/or the characters not rendering correctly.
> 
> * 938 - 952

Hey @Palipoor 

for the tasks 939, 941, 943 the output language should be `english` I have addressed the missing tags issue in #586

Issue #582, opened 2021-11-07
Task 471 examples 


danyaljj:
@Palipoor says: 

> Task 471 has very difficult instances, yet the examples are easy. 

@Maitreyapatel what do you think we move some of the instances to examples?

RushangKaria:
@danyaljj i can take this one. However im unable to assign to myself. Could you please assign it to me?

thanks!

danyaljj:
Done, thanks!

RushangKaria:
I've kept the original examples as-is, and instead appended new ones from the instances as suggested.

danyaljj:
Sounds good!

PR #596, opened 2021-11-11
Fixes Issue #582.

RushangKaria:
Kept the original examples and added a couple new ones cherry-picking from the instances.


danyaljj:
Thanks!

Issue #489, opened 2021-10-28
task 886 improvement 


danyaljj:
Based on the dataset (https://huggingface.co/datasets/quail), it looks like each paragraph comes with multiple questions. 
In which case, each instance of task 886 (question generation) should have multiple unique questions. 

FYI @saianirud

swarooprm:
@saianirud, can you fix this?

kurbster:
I can fix this issue.

kurbster:
For task 886, the Quail dataset has only one question per instance. In the schema of the dataset, the `question` value is a string, not a list. They offer 4 different answers (like multiple choice) for each question and only 1 question for each instance.

I have opened PR #598 with my changes to these tasks.

PR #598, opened 2021-11-12
Fixing issue #489. Improvements to tasks 886-889

kurbster:
Addressing issue #489. There were no additional questions in the dataset to add.

However, I shuffled the instances for the classification tasks as they were not shuffled before and I also cleared up inconsistent wording in the definitions.


kurbster:
Also, I added the proper categories and domains to each task. Except for task 889, I couldn't think of a proper domain so I left it out.

danyaljj:
Thanks!

PR #601, opened 2021-11-12
Tasks 1200-1225: Assigning categories, domains and addressing feedbacks 

yeganehkordi:


yeganehkordi:
In the atomic tasks, the crowd workers were confused by the head and tail words (head is the first phrase and tail is the second one). Do we need to change them?

swarooprm:
Categories reviewed: ready to go @danyaljj

danyaljj:
Thanks!

danyaljj:
I merged a conflict but not sure if I did it correctly. @yeganehkordi could you check. 

One question for either one of you: how is "verification" different from "detection"? We need to provide a definition for them in the document that clearly distinguishes their scopes.

yeganehkordi:
@danyaljj Thanks! We have two entity detection and abuse detection. I guess we can drop the abuse detection and entity detection out of the detection scope.
Verification is a sub-category for classification and it is used for binary classifications. For example, verify whether the claim is correct. In the verification tasks, the potential outputs are mentioned in the definition.
Detection tasks are to find something in the input. For example, detecting the entities in a review or detecting the motivation of something. Detection tasks can be text generation, classification, etc., and possible outputs are not necessarily given in the definition.

danyaljj:
Thanks! Could you clarify the distinction in the task hierarchy file?

yeganehkordi:
Sure! Done.
I'm thinking of writing documentation for the hierarchy after finalizing the categories.

danyaljj:
Thanks! Yeah, documentation would be super-helpful for any future follow-up work.

PR #603, opened 2021-11-12
Failing tests in the main branch

yeganehkordi:


swarooprm (approved):
LGTM

yeganehkordi:
@danyaljj We can change the test file to have a warning for domains instead of asserting errors before assigning all the categories and domains.

danyaljj:
> @danyaljj We can change the test file to have a warning for domains instead of asserting errors before assigning all the categories and domains.

I am happy to force these, otherwise, we might deviate from our hierarchy.

yeganehkordi:
Sounds good!

PR #607, opened 2021-11-13
Instructions fixed and human feedback addressed in mmmlu answer generation tasks (664-667, 685-737)

pulkitverma25:


danyaljj:
Thanks!

PR #625, opened 2021-11-14
Tasks 1283-1300: Assigning categories, domains and addressing feedback

yeganehkordi:
The kpa tasks need to be modified since they are almost the same. So, I didn't address the comments regarding them.


swarooprm:
Categories are good to go.

Issue #599, opened 2021-11-12
Issues with task 939, 941, 943


tanay2001:
I just noticed after #595 was merged it changed the `output_language` label for tasks `939`,`941`,`943`. The outputs for the following tasks are meant to be in English based on how the json file was created.

Just wanna confirm if this is an issue or its fine 

Thanks

danyaljj:
You're right, I missed your comment 🤦 
Let's fix these issues in a new PR.

PR #635, opened 2021-11-16
Output Language Tasks 939,941,943

tanay2001:
#599 

@danyaljj 
sorry for the delay was caught up with a submission, Thanks!


danyaljj:
Thanks!

Issue #592, opened 2021-11-10
task564_discofuse_classification


yeganehkordi:
The discourse types need to be defined in the definition of this task.

kurbster:
I can work on this.

PR #636, opened 2021-11-17
Fixing issue #592 - Updating task 564

kurbster:
Responding to issue #592. Does this suffice? Instead of defining each type individually, I described what the types stand for. For example, multiple types relate to discourse connective, which means a connective word was removed, so I just define what `CONN` means. Let me know if there's any feedback.


danyaljj:
Thanks!

Issue #645, opened 2021-11-18
Task 738 is incomplete


pulkitverma25:
Task 738 has missing perspectives for all the instances.
@Palipoor Ca you have a look once?

Palipoor:
Oh! I must've messed up in the codes somewhere. Will fix this ASAP.

danyaljj:
Good catch!

PR #646, opened 2021-11-19
Fix the missing perspectives in task 738

Palipoor:
addressed issue #645


danyaljj:
Thanks for fixing this.

Issue #577, opened 2021-11-07
Task 204 improvement 


danyaljj:
Here is one comment that we got from cloud annotators: 

> Need definitions of genres because I don't know what they all mean. **What's slate and oup?** 

I think the worker is right; the definition for some of these niche categories is not quite clear.

danyaljj:
@aarunku5 wondering if you can help with addressing issue?

danyaljj:
@yeganehkordi 's comment on this task: 

> In task 204, based on the paper, genres refer to the source of collecting the premises. I don't think if the classification of the sentences based on the genres is practical. For example, two-sided, telephone conversations that took place in 1990 or 1991 (TELEPHONE) and Collection of two-sided, in-person conversations that took place in the early 2000s (FACE-TO-FACE) are hard to distinguish.

yeganehkordi:
Also, tasks 197 and 198 have the same problem with the genre.

RushangKaria:
@danyaljj i can take this on!

danyaljj:
@RushangKaria checking if there are any updates on this thread?

RushangKaria:
@danyaljj Will put up the PR tomorrow.

PR #649, opened 2021-11-20
Fixes Issue #577.

RushangKaria:
resolves issue #577


danyaljj:
Thank you!

danyaljj:
One minor issue: could you put the genre names inside a single quote so that they're not confused with the rest of the text?

RushangKaria:
Yeah, I'll add that change in an hour or so.

danyaljj:
Thank you!

Issue #631, opened 2021-11-15
mmmlu tasks improvement 


danyaljj:
Our human annotators are complaining about the task definitions for the MMMLU tasks. 

I think part of the issue is that the tasks don't explain enough about the desired outputs. 
For example, here: 
https://github.com/allenai/natural-instructions-expansion/blob/master/tasks/task667_mmmlu_answer_generation_business_ethics.json

The only thing said is that `"You are given a question on business ethics.` But then: How do the expected answer relate to `business ethics`? (should they pick something that is more ethical or less ethical?) What is even "business ethics", to begin with? We need to improve the instructions to address these.

danyaljj:
@Sujan242 could you work on improving these task definitions?

danyaljj:
@Sujan242 checking there are any updates on this thread.

Sujan242:
Hey Daniel! 
I have my university exams going on now. Is it fine if I push my changes by Wednesday?

danyaljj:
Sounds good.

Sujan242:
@pulkitverma25 are there any updates on Discussion #467. The conflicts might be difficult to resolve if we're working on the same tasks.

Sujan242:
> I think part of the issue is that the tasks don't explain enough about the desired outputs. For example, here: https://github.com/allenai/natural-instructions-expansion/blob/master/tasks/task667_mmmlu_answer_generation_business_ethics.json
> 
> The only thing said is that `"You are given a question on business ethics.` But then: How do the expected answer relate to `business ethics`? (should they pick something that is more ethical or less ethical?) What is even "business ethics", to begin with? We need to improve the instructions to address these.

I have added the definitions for lesser-known subjects like business ethics in my PR. However, things like "should they pick something that is more ethical or less ethical?" are quite clear when you are looking at an individual instance. Let me know if this is fine or it should be improved further.

pulkitverma25:
> @pulkitverma25 are there any updates on Discussion #467. The conflicts might be difficult to resolve if we're working on the same tasks.

Since the updates are in different parts of the files, this shouldn't be difficult. You go ahead and submit your changes, I will take care of the merge.
To avoid any confusion: The only issue related to MMMLU tasks that I am currently working on is #569

PR #654, opened 2021-11-24
MMMLU task improvements

Sujan242:
Fixes Issue #631. 
#607 addresses most of the human feedbacks for these tasks. 
I have added definitions of some of the lesser-known subjects.


danyaljj:
Thanks!

PR #659, opened 2021-11-27
Tasks 1300-1325: Assigning categories and domains 

yeganehkordi:


swarooprm (approved):
Merging this as it does not contain addressing human feedbacks.

PR #660, opened 2021-11-29
Tasks 1325-1350: Assigning categories and domains

yeganehkordi:


swarooprm (approved):
Category and domains are good to go @danyaljj

PR #669, opened 2021-12-05
fixed explanation of task456 (fixes #581)

### LINKED ISSUE #581: Task 456 improvement
@Palipoor says: 

> Task 456 was difficult. I think I didn't understand it either.

Here is the actual feedback: 

> regarding instructions: You should emphasize that the verb in question is the one in parenthesis- not any other potential intentions within the sentence. Also define "unconditional" or provide an example. The HITs should have a yes or no bubble, and an extra fill in the blank if an explanation is needed or desired.
 > regarding p examples: I understood these easily. I'd have liked to see something "conditional" to know what that meant.
 > regarding n examples: I didn't understand these and would have answered the opposite of what was provided. I'd need a few more examples to make sense of them.

pulkitverma25:


danyaljj:
Thanks!

PR #676, opened 2021-12-10
Fix #488. Marked all pronouns in all instances.

### LINKED ISSUE #488: task 646 improvement
Leftovers from PR#306:  
There are many examples where there is more than one pronoun in the sentence.
Like here, the gold output should be ["it", "he"]. 
For having a more reliable evaluation, I think we should include all these valid answers in the output. 

FYI @khushal1996

pulkitverma25:


tasks/task646_answer_generation.json
```diff
@@ -30,20 +30,20 @@
...
+            "output": "he, him",
             "explanation": "here, the input statement talks about ethan running and referring to him as he"
         }
     ],
     "Negative Examples": [
         {
             "input": "madonna fired her trainer because she couldnt stand her boyfriend.",
             "output": "his",
-            "explanation": "there is no his in the sentence"
+            "explanation": "there is no his in the sentence. The correct answer should be 'her'."
```
Palipoor: 'her' and 'she'.

tasks/task646_answer_generation.json
```diff
@@ -30,20 +30,20 @@
...
         },
         {
             "input": "this book introduced shakespeare to ovid; it was a fine selection of his writing.",
             "output": "it",
-            "explanation": "the pronoun here is referring to ovid"
+            "explanation": "there are two pronouns in the sentence, the correct answer should be 'it, his'."
```
Palipoor: I think 'this' is also a pronoun.

Palipoor (changes requested):
The examples(and the instructions) here do not match the instances. In the instances, each acceptable answer (i.e. each member of the list of outputs) contains only one pronoun, whereas in the examples the given acceptable answer contains all the pronouns.

pulkitverma25:
> The examples(and the instructions) here do not match the instances. In the instances, each acceptable answer (i.e. each member of the list of outputs) contains only one pronoun, whereas in the examples the given acceptable answer contains all the pronouns.

Many of the instances  (almost half of them) contain multiple pronouns.

pulkitverma25:
Also, the tests are failing because they check for skewness in the number of appearances of individual pronouns. I think we should skip checking for skewness for this task specifically.
Is this ok, @danyaljj ?

tasks/task646_answer_generation.json
pulkitverma25: Fixed

tasks/task646_answer_generation.json
pulkitverma25: Yes, I missed demonstrative pronouns. Added them now.

Palipoor:
> Many of the instances (almost half of them) contain multiple pronouns.

What I am saying is this:

In the examples, we have : 
```
        {
            "input": "the user changed his password from grwqwu8jyc to\u00c2 willow-towered canopy huntertropic wrestles as it was easy to remember.",
            "output": "his, it",
        },
```  
So here the accepted output is "his, it". 
but, in the instances, we have this:
```
        {
            "input": "the lawyer asked the witness a question, but he was reluctant to repeat it.",
            "output": [
                "he", "it"
            ]
        },
```
This means for this input, we have two different possible answers. Either "he" or "it". If the task is to list all the pronouns, for the above instance the output should be:
```
        {
            "input": "the lawyer asked the witness a question, but he was reluctant to repeat it.",
            "output": [
                "he, it",
                "it, he"
            ]
        },
```  
Or for simplicity, you can point out that the pronouns should be in the same order as the text, so the only acceptable answer would be "he, it".
Does it make sense?

danyaljj:
@pulkitverma25 just checking if you have seen @Palipoor message above.

pulkitverma25:
@Palipoor I think it would make sense to ask for pronouns in the same order as their first appearance. I'll add more examples to make this point clear. The other option will involve giving all permutations of pronouns in a sentence as acceptable answers, which I think would be unnecessary.

@danyaljj Thanks for tagging. I completely missed this.

danyaljj:
> @Palipoor I think it would make sense to ask for pronouns in the same order as their first appearance. 


If this order is explicitly stated in the instruction, that'd be fine.

danyaljj:
@Palipoor can you check if the PR looks good to you?

Palipoor:
Hi,
this has to change:
in the instances, the outputs are like:
`"output" : ["he", "him"]`  
they should be like this:
`"output" : ["he, him"]`

pulkitverma25:
@Palipoor Please have a look at it now. I hope everything is fixed.

Palipoor:
LGTM! Thanks!

pulkitverma25:
@danyaljj This is ready to be merged in master.

Issue #652, opened 2021-11-23
Task1412_dart_predicate_identification skew


yeganehkordi:
This task seems to be heavily skewed toward `food`. 
@danyaljj Shall we consider re-sampling or are we going to solve it with down-sampling later?

danyaljj:
It is skewed, but it's not too bad. 
We should fix it, but definitely not a priority.

yeganehkordi:
Thanks! Also, I guess it's not present in the Tasks statistics file because it has more than 15 outputs. Maybe we can consider a higher threshold.

danyaljj:
That's a good observation! We should consider increasing that limit.

yeganehkordi:
We may need to set the limit higher than 45. I've checked this task now. Here is the result:
   📋 classes: ['1880' '1913' '1960' '1965' '1968' '1970' '1975' '1977' '1985' '1992'
 '2.20' '2000' '2002' '2007' '2008' '2009' '2010' '2011' '2013' 'B.P.'
 'CAN' 'GP' 'HR' 'IATA' 'IPA' 'L3' 'MVP' 'N/R' 'NC' 'PF' 'PLACE' 'RF'
 'SCORE' 'SPG' 'SURFACE' 'TDP' 'TFR' 'V2' 'V3' 'V4' 'customer rating'
 'food'] 
   📋 Norm_counts: [3.38409475e-04 3.38409475e-04 3.38409475e-04 3.38409475e-04
 3.38409475e-04 3.38409475e-04 6.76818951e-04 3.38409475e-04
 6.76818951e-04 3.38409475e-04 3.38409475e-04 3.38409475e-04
 3.38409475e-04 6.76818951e-04 6.76818951e-04 1.35363790e-03
 3.38409475e-04 3.38409475e-04 6.76818951e-04 3.38409475e-04
 3.38409475e-04 3.38409475e-04 3.38409475e-04 3.38409475e-04
 3.38409475e-04 6.76818951e-04 1.35363790e-03 3.38409475e-04
 3.38409475e-04 6.76818951e-04 3.38409475e-04 3.38409475e-04
 3.38409475e-04 3.38409475e-04 3.38409475e-04 3.38409475e-04
 3.38409475e-04 3.38409475e-04 3.38409475e-04 3.38409475e-04
 3.72250423e-03 9.78341794e-01] 
   📋 Distribution of classes: [   1    1    1    1    1    1    2    1    2    1    1    1    1    2
[...]

PR #680, opened 2021-12-13
Fix for Issue #652.

RushangKaria:


danyaljj:
I'm not sure if I understand the changes in the PR. 
Isn't a predicate supposed to be "part of a sentence"? If so, what does "NAME" mean in the following instance? 

```
        {
            "input": "The route 42, 43 are Mount pleasant line.",
            "output": [
                "NAME"
            ]
        },
```

RushangKaria:
Copying over the comment i made in issue #652.

> Hi @danyaljj

> It looks like the "food" label is the only predicate that actually appears in the sentence.
> Other sentences in this dataset have different special labels that appear as predicates for which there is no documentation.

> I've put up a PR using these special tags however im not sure where to find their definitions.

For the predicates that actually appear in the sentences, "food" appears in most of the cases as observed by the skew. If i remove the code that checks whether the predicate actually appears in the sentence, then the predicates have these special tags which were added by the annotators for the predicates.

Not sure where the definitions for them lie.

danyaljj:
Got it. Let's remove this task.

RushangKaria:
Cool! Should i delete both the file and the entry in README.md?

danyaljj:
Yes

Issue #611, opened 2021-11-13
task 27 skew 


danyaljj:
DROP classification into (Date.|Number.|Span.) is heavily skewed.
@swarooprm do you happen to have your scripts for this dataset?

Palipoor:
I will fix this.

PR #681, opened 2021-12-14
Fix the imbalanced labels in task 27

Palipoor:
Solves #611


danyaljj:
Awesome, thanks! Could you also shuffle the instances?

Issue #579, opened 2021-11-07
task383_matres_classification


danyaljj:
According to @Palipoor: 

> task383_matres_classification was difficult for annotators to understand. I didn't understand it either!

Here are the actual feedback: 

> regarding instructions: The word "anchored" didn't click with me. The phrasing on the instructions sounds much more confusing than the task. Also, fix the line break issue after "verb separated with a"<newline><newline>Most importantly, do I just type "yes" or "no"?<newline><newline>Switching from one HIT that was "remove words of 7 letters" to this task is very confusing and overwhelming. They should be different batches.<newline><newline>The time it takes to read the instructions and do the task is much longer when they are intermixed like this and end up being underpaid or make me want to return. I'm already over 13 minutes on this one!<newline>
> regarding p examples: I definitely need more examples. Why isn't "invite" anchorable? It is a verb and it happened in the past because they decided to invite at last years (in the past) convention. Or is it only anchorable if it was "invited". This is really confusing!
 > regarding n examples: I liked that the first example was longer and the second shorter. But, I think we need a few more long examples where the word doesn't come at the end, which makes me want to skim.

yeganehkordi:
They are right! This task is hard. I'll try to improve the definitions and the examples.

pulkitverma25:
Same issue with tasks 456, 457, 458, and 459.
They have similar descriptions

.

danyaljj:
@yeganehkordi if this still an issue or we should close it?

yeganehkordi:
Sorry for being late. I created the PR, and I think it's not an issue anymore.

PR #684, opened 2022-01-04
Matres tasks enhancement

yeganehkordi:
Addressing #579


tasks/task457_matres_conditional_classification.json
```diff
@@ -11,18 +11,23 @@
     "Domains": [
         "News"
     ],
-    "Definition": "You will be given a context and a verb separated with a \n, and you have to answer if the given verb implies a hypothetical, or condition or not. We say a verb is hypothetical if it involves or is based on a suggested idea or theory. A verb is conditional if something is essential to the appearance or occurrence of something else. The output should be \"Yes\" if the verb is conditional or hypothetical and \"No\" otherwise.",
+    "Definition": "You will be given a context and a verb separated with a newline character, and you have to answer if the given verb implies a hypothetical, or condition or not. We say a verb is hypothetical if it involves or is based on a suggested idea or theory. A verb is conditional if it is in conditional sentences and something is essential to the appearance or occurrence of something else. Note that an event that happened in the past or it is happening now cannot be conditional. The output should be \"Yes\" if the verb is conditional or hypothetical and \"No\" otherwise.",
```
Palipoor: This definition can be worded better. For example:
`You will be given a context and a verb separated with a newline character. You have to decide if the given verb implies a hypothetical or conditioned action or not. We say a verb is hypothetical if it involves or is based on a suggested idea or theory. A verb is conditional if it is in a conditional sentence, and its occurrence relies on something else. Note that an event that happened in the past or is happening now cannot be conditional. The output should be \"Yes\" if the verb is conditional or hypothetical and \"No\" otherwise.`

tasks/task456_matres_intention_classification.json
```diff
@@ -11,7 +11,7 @@
     "Domains": [
         "News"
     ],
-    "Definition": "You will be given a context and a verb separated with a \n, and you have to identify if the given verb implies eiher an opinion, an intention, or a wish. The output should be: 'Yes' if the verb irefers to an opinion, an intention, or a wish; and 'No' otherwise. Please note that a verb refers to an intention only if it refers to an event planned to happen in the future and is not conditional (or part of a condition).",
+    "Definition": "You will be given a context and a verb separated with a newline character, and you have to identify if the given verb implies eiher an opinion, an intention, or a wish. Please note that a verb refers to an intention only if it refers to an event planned to happen in the future and is not conditional (or part of a condition). The output should be: 'Yes' if the verb irefers to an opinion, an intention, or a wish; and 'No' otherwise. ",
```
Palipoor, line 14: The definition can be worded better (the way it is now, can be vague and interpreted as if we have multiple classes: intention, opinion, wish, ...)
Something like this:
`You will be given a context and a verb separated with a newline character. You have to identify verbs that imply an opinion, an intention, or a wish. Please note that a verb refers to an intention only if it refers to an event planned to happen in the future and is not conditional (or part of a condition). The output should be: 'Yes' if the verb refers to an opinion, an intention, or a wish; and 'No' otherwise.`

Palipoor (changes requested):
Looks good to me, other than the two comments on the definitions.
Also, there are some " `` " characters here and there. I think it's better if we replace them with backslash + ".

yeganehkordi:
I updated the definitions and punctuations. I realized those punctuations were redundant, so I dropped them.

PR #689, opened 2022-01-16
Drop links from TriviaQA

yeganehkordi:
- Dropping the links from TriviaQA tasks
- Update the test all to detect URLs in the tasks

### DIFF
--- a/src/test_all.py
+++ b/src/test_all.py
@@ -213,7 +213,7 @@ for file in files[begin_task_number:end_task_number + 1]:
                                         f"Merge outputs before removing duplicates. :-/ \n {instance}")
                                         
             # Make sure there are no link in instances
-            url_reg  = r'[a-z]*[:.]+\S+'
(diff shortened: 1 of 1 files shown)


src/test_all.py
```diff
@@ -211,7 +211,16 @@ def natural_keys(text):
...
-
+                                        
+            # Make sure there are no link in instances
+            url_reg  = r'http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+'
+            instances = data['Instances']    
+            for instance in instances:
+                ck_url = re.findall(url_reg, instance['input'])
+                if ck_url:
+                    print(f'⚠️ WARNING: Looks like there is a link in the input: {ck_url}')
+                    break
```
danyaljj, line 222: should this raise an exception (rather than `break`)?

danyaljj:
Thanks! Left a question above.

src/test_all.py
yeganehkordi: We have many tasks with links in the input and we need to change some of their definitions. After updating them, I'll change this scope to raise an exception.

danyaljj: I see. Let's remember to address these definitely in near future.

PR #691, opened 2022-01-18
Final version of task-hierarchy

yeganehkordi:
Some thoughts on hierarchy:
- We have many `Incorrect some-category` categories. We can add an `Incorrect generation` category and use `Incorrect generation` + `some-category` instead.
- I think `Classification -> Verification -> Claim Verification` is same as `Classification -> Verification`.
- We have `Paraphrase/Translation Verification`, `Paraphrasing/Translation`, and `Paraphrase/Translation Generation` which are redundant.
- If we are going to drop the links, we can drop `Text Generation -> Weblink Generation` category.
- I think `Sorting` and `Order generation` categories are the same.
- What is `Relevancy Estimation`?
- `Relation Prediction` and `Detection -> Relation Detection` are almost the same.

### DIFF
--- a/doc/task-hierarchy.md
+++ b/doc/task-hierarchy.md
@@ -9,7 +9,6 @@
 ## Task Categories   
 - `Classification`
   - `Classification -> Answer Type Classification`: Classify a text that contains a question to the type of answer that is expected (e.g., multiple choice, extractive, abstractive etc.) 
-  - `Classification -> Incorrect Classification`
   - `Classification -> Intent Classification`: Classify a text to a pre-defined set of intent classes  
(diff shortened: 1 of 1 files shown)


danyaljj:
Thanks! Looks good to me! 
I have a limited understanding of the category names at this point, but the suggestions sound reasonable. @swarooprm might have better suggestions here. 


For future PRs: we should revise the tasks so that we don't get the "⚠️ WARNING: did not find" warnings anymore.

danyaljj (approved):

swarooprm:
@yeganehkordi, merging based on our discussion in the meeting.

swarooprm (approved):

PR #698, opened 2022-02-05
Add examples to task 697 and modify some instances to fix #661

### LINKED ISSUE #661: task697 is incomplete.
For any question on formal logic sentence, vocabulary, universe, and semantics should also be given to validate if the sentence evaluates to true or false. The task assumes an implied mental model of these things. For any statement, I can create a universe, vocabulary, and semantics to make all options true. I think we should remove this task.

Palipoor:
(I am sorry that the change is too big. I had to fix the Unicode characters so I could understand the logic notation.)
I added a couple of examples to the task and modified some of the instances. In most cases, these kinds of instances were like this:  
`            "input": "Select the best translation into predicate logic. David teaches Chris. (c: Chris; d: David; Txy: x teaches y)\n(A)Tdc (B)Tcd (C)Tcc (D)dTc",
`  
But in some cases, the symbols weren't defined. I updated those with symbol definitions.

### DIFF
--- a/tasks/task697_mmmlu_answer_generation_formal_logic.json
+++ b/tasks/task697_mmmlu_answer_generation_formal_logic.json
@@ -18,9 +18,19 @@
             "explanation": "Because of the reasons specified in the argument, it is difficult to convince people of the unconscious, and in turn to teach them what their conscious knowledge contradicts."
         },
         {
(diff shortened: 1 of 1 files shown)


danyaljj:
Thanks!

Issue #685, opened 2022-01-06
Task1562_clickbait_new_bg_answer_generation


yeganehkordi:
This task is to generate the fake news score based on the given news.
`Fake-news score is a numeric value indicating whether the given news article is fake or not. The score is a numeric value ranging from 0 to 10.`
I think it can be so hard even for humans to generate this score because we don't have a metric for recognizing the difference between the scores. I saw we don't have feedback on this task, and I guess it's because the task is in Bulgarian.

task1563_clickbait_new_bg_answer_generation has the same problem. 
`You are expected to generate a value for the click-bait score. Click-bait score is a a numeric value indicating whether the given article is click-bait or not. The output can contain only positive numeric values of upto 10 digits in length.`

danyaljj:
Agreed. I am fine if we drop these tasks.

Issue #452, opened 2021-10-15
Two tasks numbered 119!


Palipoor:
Hi!
For some reason we have two tasks numbered 119.

eshaanpathak:
Yeah, I pointed this out in #224. Daniel said he was going to renumber tasks at the end once everything else is finalized.

swarooprm:
We also have two tasks numbered 868

aarunku5:
There are also two tasks numbered 672

pulkitverma25:
@danyaljj Are we going to renumber all the tasks?
If not, I can simply renumber the repeating task numbers.

danyaljj:
Thanks! We should definitely address these and have a test that prevents repeated numbers.

PR #699, opened 2022-02-05
Drop 1562 and 1563 because of vagueness reported in #685

Palipoor:


danyaljj:
Thanks! For future PRs, lets fill up this gap with the tasks discussed in #452?

PR #702, opened 2022-02-05
Fix the imbalance issue in instances

Palipoor:
The original dataset had only ~ 850 instances labeled 1, so I downsampled the 0 ones. 

Also, relevant to this:
Task 839 is very weird. It's claimed to be driven from the same dataset(cdt). At first, I thought it was the same task as 838, only with "yes" and "no" as labels. Then I realized the neg and pos labels are balanced in this task. I checked the dataset repository and it seems like they have another task for "detecting if a tweet is cyber-bully or hate speech or neither" which is itself very skewed and I don't think we should add it. So I thought maybe that's the task 839, but it wasn't. I was wondering where all these yes-labeled tweets have come from and I realized it's the same data, but every "yes" labeled instance has a " Question: Is the tweet free of any cyberbullying (harmful) content?" at the end of it, and every "no" labeled instance has a "Question: Does the tweet contain cyberbullying (harmful) content?" at the end of it! Hence artificially making half of the labels "No" and half of them "Yes". I think we should drop task 839.


danyaljj:
Thanks! Can you please drop it from this exception list? https://github.com/allenai/natural-instructions/blob/master/src/test_all.py#L120

Palipoor:
@danyaljj Done.
What do you think about task 839?

danyaljj:
if they're the same task (are they?), how about we merge them? Let's move this discussion to a separate issue since its orthogonal to this PR.

PR #704, opened 2022-02-07
Task050 label imbalance bug fix

swarooprm:
Positive labels:3250
Negative labels: 3250

### DIFF
--- a/src/test_all.py
+++ b/src/test_all.py
@@ -117,7 +117,7 @@ if not args.task:
 
 # TODO: over time, we need to fix the skew of the following tasks
 skew_exclusion = [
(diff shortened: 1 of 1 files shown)


swarooprm:
Fixing https://github.com/allenai/natural-instructions/issues/619

danyaljj:
Thanks! Can you drop the task from the exception list? 
https://github.com/allenai/natural-instructions/blob/master/src/test_all.py#L120

swarooprm:
> Thanks! Can you drop the task from the exception list? https://github.com/allenai/natural-instructions/blob/master/src/test_all.py#L120

done.

PR #705, opened 2022-02-08
Revising remaining tasks

yeganehkordi:


swarooprm (commented):
I checked the first few files and the changes (mostly addition of a category or renaming) makes sense, even though there is an indentation issue, however it seems like some files were deleted (e.g. files with scruple task). I am not able to verify this because the diff is too large to view.

Can we ensure that we are not changing any other content apart from the categories? @yeganehkordi

yeganehkordi:
I guess the indentation issue is because of github preview, atomic tasks look good on my system. However, some tasks didn't have a good indentation and I assumed we are going to fix them later.
I dropped link generation tasks and tasks related to fake news score intentionally. I'll check the commits now to make sure I'm not changing anything else.

yeganehkordi:
@swarooprm They're fine. I checked them and deleted files (e.g., files with scruple tasks) were .bak files.

swarooprm:
> @swarooprm They're fine. I checked them and deleted files (e.g., files with scruple tasks) were .bak files.

Thanks for confirming. The changes look good, but I have not verified all files yet, not able to load them all in github diff.

yeganehkordi:
> Thanks for confirming. The changes look good, but I have not verified all files yet, not able to load them all in github diff.

I think you can load the files in each commit separately.

swarooprm:
Changes look reasonable to me, the categories look more consistent than before.

Issue #589, opened 2021-11-10
Task547_spl_translation_entk_en and task560_alt_translation_en_entk


yeganehkordi:
The instruction is incomplete, and it seems the task is to remove the space before punctuation marks. @PhaniRohithaKaza Can you explain the tokens?

PhaniRohithaKaza:
@yeganehkordi The task is sentence translation and the task what usually does is convert given english sentence to it's tokens. But as tokens are also in english the input and output doesn't differ. Even I'm confused on this task. @swarooprm can you please look into it and help us.

yeganehkordi:
Yeah, except for punctuations, they seem to be the same.

Palipoor:
Yeah, "English tokens" is one of the translated versions in their data. I think we can drop these two tasks. @danyaljj 
Also, some of the other tasks from this dataset have two "Domains". I can fix these in a PR.

danyaljj:
Sounds good, thank you!

danyaljj:
Moving some of the comments from #709 to here: 

@yeganehkordi 's comment: 
> I'm in favor of not dropping en_entk and entk_en tasks. We already have simpler tasks than these tasks. I think we can change their definition and keep them. 

@Palipoor 's comment: 
>  I think those tasks being simple(just putting space before punctuation marks or removing the space before them) is one thing, but the other thing is that the models are probably going to process these at the token level. So it seems like both the input and the output are going to end up being encoded the same way, which makes the task pointless.

swarooprm:
In my opinion, we could keep en_entk and entk_en tasks and mention in the definition that the task is to remove/add space before punctuation. This is a simple task, but not an invalid task.

Let's not worry about how models are going to process it. E.g. even if the encoder ignores the space, the decoder has to decode it back with space. 

My opinion is not very strong and I am also fine if we decide to delete it. However, we should note that, creating a task requires significant effort, and deleting is easy. So, we should avoid deleting as much as possible and think about repairing.

yeganehkordi:
I think in the worst-case scenario, we can shuffle the tokens in the input and change these tasks to "order generation" tasks.

PR #709, opened 2022-02-09
Remove duplicate and incorrect domains for alt tasks and drop two tasks

Palipoor:
I removed the "Public Places" domains and changed them to "News", as all the tasks were news. (They must've been mixed up with spl tasks?)  
I also dropped the en_entk and entk_en tasks, as discussed in #589.


yeganehkordi:
Thanks, removing the duplicate domains has been addressed in #705.
Also, I'm in favor of not dropping en_entk and entk_en tasks. We already have simpler tasks than these tasks. I think we can change their definition and keep them.

danyaljj:
@Palipoor let's get this PR updated based on the conflicts? 
Also, seems like we need to discuss #705 a bit more, given @yeganehkordi 's comment above.

Palipoor:
@danyaljj Yes. I think those tasks being simple(just putting space before punctuation marks or removing the space before them) is one thing, but the other thing is that the models are probably going to process these at the token level. So it seems like both the input and the output are going to end up being encoded the same way, which makes the task pointless.

PR #711, opened 2022-02-10
remove URLs from the tasks

yeganehkordi:
- I dropped the Reference class from task375. `This class refers to a reference to a source of objective evidence. It is typically a citation or a URL of a published work from a renowned source`. There was only 17 instances from this class and most of them were URLs. By removing the URLs, instances were meaningless.

### DIFF
--- a/tasks/task115_help_advice_classification.json
+++ b/tasks/task115_help_advice_classification.json
@@ -11268,12 +11268,6 @@
             ]
         },
         {
-            "input": "]",
-            "output": [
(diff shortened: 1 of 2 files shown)


danyaljj:
Thanks @yeganehkordi! 

For future PRs: let's keep them PRs relatively small so that it's easier to review.

yeganehkordi:
Sure! I'll use another PR for the rest of the work.
This PR is ready to merge.

PR #712, opened 2022-02-13
Adding reasoning field - tasks1-50

yeganehkordi:
I wrote a script for separating the reasoning field in the tasks and updated the first 50 tasks. I'll update all the files in the next PRs.

### DIFF
--- a/tasks/task003_mctaco_question_generation_event_duration.json
+++ b/tasks/task003_mctaco_question_generation_event_duration.json
@@ -10,7 +10,31 @@
         "Question Generation -> Contextual Question Generation",
         "Question Generation -> Commonsense Question"
     ],
+    "Reasoning": [
+        "Temporal Reasoning",
+        "Commonsense Reasoning"
+    ],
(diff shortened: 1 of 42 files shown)


src/auto_add_reasoning.py
```diff
@@ -0,0 +1,42 @@
+import json
```
danyaljj: if we're adding the reasoning categories to the files, why do we need to check in this file?

danyaljj (approved):
One small suggestion; otherwise, everything looks good.

src/auto_add_reasoning.py
yeganehkordi: Makes sense :) I'll drop it now.

swarooprm:
All the PRs on reasoning field: Is it about separating the categories with the word 'reasoning' and storing in a different field? If this is the case, we might not need to review. @yeganehkordi

yeganehkordi:
> All the PRs on reasoning field: Is it about separating the categories with the word 'reasoning' and storing in a different field? If this is the case, we might not need to review. @yeganehkordi

Yes, I've separated the reasoning field and sorted the keys.

swarooprm (approved):

PR #719, opened 2022-02-15
Tasks1385-1390: ANLI, CB, HellaSwag, WSC 

yeganehkordi:
I'm adding the missing evaluation tasks.

### DIFF
--- a/tasks/task1385_anli_r1_entailment.json
+++ b/tasks/task1385_anli_r1_entailment.json
@@ -15,7 +15,7 @@
 		"Commonsense Reasoning",
 		"Causal Reasoning"
     ],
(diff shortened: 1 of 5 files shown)


Palipoor:
Hi! These look good to me. Just a few small notes:
1. Instead of `Classify your answers into \"Contradiction\", \"Neutral\", or \"Entailment\"."` maybe we can say `answer with "Contradiction", "Neutral", or "Entailement" `. The same thing with the hallaswag and wsc definitions.
2. In `task1386_anli_r2_entailment`: cannot -> can not
3. In the hellaswag task, I think the second positive example needs a bit more clarification in the explanation.
4. In the wsc task, it's mentioned that the pronoun is in brackets, but it's inside parentheses.

yeganehkordi:
Hey, Thanks!
I've addressed your comments. Although I think cannot is correct :)

Palipoor:
Yeah, my bad!

PR #722, opened 2022-02-20
Script for updating categories, domains, and reasoning

yeganehkordi:


doc/categories.json
```diff
@@ -0,0 +1,580 @@
+{
```
danyaljj: It might be better to have a more informative name for this file, such as `categories_map.json`.

doc/domains.json
```diff
@@ -0,0 +1,347 @@
+{
```
danyaljj: Same comment as above.

doc/reasonings.json
```diff
@@ -0,0 +1,125 @@
+{
```
danyaljj: Same comment as above.

src/update_categories.py
```diff
@@ -0,0 +1,66 @@
+import json
+from os import listdir, path
+from os.path import isfile, join
+import argparse
+import re
+
+"""
+Script for updating the categories, domains, and reasoning field for all the tasks.
```
danyaljj: "update" is a bit open-ended. How about "simplify" or "sparsify"? 
Say a few words as to why we're doing this.

danyaljj (approved):
Few small suggestions. Otherwise, looks great! 💯

yeganehkordi:
Thanks! Done.

PR #730, opened 2022-02-25
Winogrande tasks

yeganehkordi:
I removed the context word in task 33 and added a new Winogrande task with options.
@yizhongw @danyaljj


danyaljj:
Thanks!

Issue #626, opened 2021-11-14
Tasks 1285-1307: kpa dataset


yeganehkordi:
These tasks have the same definitions with different domains. @ashok-arjun, Can you change the definitions? For example, we can have summary generation, argument generation, etc.

mishra-sid:
Hi, since it requires just updating the task definitions, I can work on this.

yeganehkordi:
Hi, it's not just the definition; the whole task needs to be changed. These tasks are the same, and we need to change some of them to have new different tasks.
It would be great if you could work on this.

danyaljj:
@yeganehkordi Could you elaborate more on what is the issue?

yeganehkordi:
Yeah, the kpa dataset consists of arguments and their summary in different topics(gathered from many sources). Each of these tasks is to verify the summary in a given topic, and the only difference in the tasks is their topic. Actually, I think the contributor didn't need to separate the topics into different tasks, because knowing the topic doesn't make any difference in summary verification. 
My suggestion is to provide new tasks based on this dataset or dropping some of the tasks. The new tasks can be summary generation, argument generation, topic verification, etc.

mishra-sid:
Makes sense, thanks for the clarification. Maybe I can drop all the existing datasets on kpa, aggregate the instances from those datasets and make the new suggested tasks ?

yeganehkordi:
I agree! I guess it would be preferable to provide as many different tasks as you can.

danyaljj:
Yeah, if it's too difficult to fix a task, it's better to drop it completely.

mishra-sid:
@danyaljj Do you think it's worth spending time creating new tasks using this data(maybe just the simple ones?) or should I just create a PR to drop the existing tasks?

danyaljj:
Let's drop it.

danyaljj:
Actually, looking through the conversation, not sure why I suggested to drop the tasks. How about we go ahead with @yeganehkordi 's suggestion of merging the tasks? (if I understood it correctly). @Palipoor do you have any suggestions here?

ashok-arjun:
Sorry I have been busy over the past few months, and couldn't be active here.

Actually, I think that having the topic (eg. `Assisted suicide should be a criminal offence`) makes a lot of difference in a specific task, as it covers a part of the question for all the instances. I feel it is similar to having the topic as `translate from en-->fr` and then having instances under the task. 

[...]

PR #733, opened 2022-03-01
Merge all kpa tasks

Palipoor:
Due to the discussion under #626


yeganehkordi:
Thanks! I think we don't need the topic in the input.
Also, I was thinking if we can keep them and evaluate generalization over domains using these kinds of tasks.

danyaljj:
Sorry I missed the PR. Just seeing it now.

yizhongw:
I think we certainly don't want to have different topics as separate tasks (they are too many!). For domain generalization evaluation, we probably don't want to evaluate using just one task but on different domains. So, I think merging all the topics into one task is fine.

On whether to put topic in the input, I think either way is fine. I vote for keeping them, since this information is provided in the original task and we probably don't want to make the assumption that this information is not needed.

danyaljj:
@yeganehkordi do you have any strong opinions regarding @yizhongw 's comment above?

yeganehkordi:
> @yeganehkordi do you have any strong opinions regarding @yizhongw 's comment above?

No, it sounds good! My opinion is not strong and I'm fine if we decide to merge them and keep topics.

PR #734, opened 2022-03-08
Add script to covert "prompt-source" tasks

yeganehkordi:
I wasn't sure if we need to add other keys such as categories or not. I've added them in the script for consistency with our tasks.


danyaljj:
Thanks! Merging this for now. Let's add the other details (shuffle + subset selection) in a different PR.

PR #736, opened 2022-03-10
Update convert_prompt.py

yeganehkordi:
I've updated the code. Please let me know if I missed sth.
changes:
- Adding subset selection for each dataset
- Adding prompt name and id to the json files

### DIFF
--- a/src/convert_prompt.py
+++ b/src/convert_prompt.py
@@ -39,15 +39,19 @@ if args.subset:
 
 def get_dataset(dataset_name):
     dataset = load_dataset(dataset_name, split="train")
+    cap = min(6500, len(dataset))
+    samples = random.choices(dataset, k=cap)
     # Load prompts for this dataset
     dataset_prompts = DatasetTemplates(dataset_name)   
-    return dataset, dataset_prompts
+    return samples, dataset_prompts
 
 def get_subset(dataset_name, subset_name):
     dataset = load_dataset(dataset_name,subset_name, split="train")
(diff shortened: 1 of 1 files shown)


src/convert_prompt.py
```diff
@@ -39,15 +39,19 @@
 
 def get_dataset(dataset_name):
     dataset = load_dataset(dataset_name, split="train")
+    cap = min(6500, len(dataset))
```
danyaljj: is it possible to make this an optional parameter? 
Basically, we don't want to sample by default, unless `--sample 6500` is given in the input.

yeganehkordi: Ya, I'll fix it now.

yizhongw:
This conversion script looks good to me!

yizhongw:
Question: @yeganehkordi where are you going to put the shared task lists between NI and PromptSource?

yeganehkordi:
> Question: @yeganehkordi where are you going to put the shared task lists between NI and PromptSource?

I'll update the spreadsheet now and list the shared tasks there.

danyaljj:
> I'll update the spreadsheet now and list the shared tasks there. 

It might be easier if you share the list as a json file in a different PR. This way, any followup work can also use this mapping for conducting further research.

yeganehkordi:
> > I'll update the spreadsheet now and list the shared tasks there.
> 
> It might be easier if you share the list as a json file in a different PR. This way, any followup work can also use this mapping for conducting further research.

Sounds good! Will do.

PR #737, opened 2022-03-11
Create prompt_task_map.json

yeganehkordi:
Here is the map of the shared task between our tasks and training datasets of the T0 model.
They have trained the model on the 35 datasets. We have 31 shared tasks and lots of missing tasks that I've listed in the spreadsheet.
Also, for the paws, duorc, amazon_us_reviews, and hotpotqa datasets, our tasks don't have a specific subset. So, we may need to add them again.

### DIFF
--- a/doc/prompt_task_map.json
+++ b/doc/prompt_task_map.json
@@ -1,31 +1,83 @@
 {
     "commonsense_qa": [
-        "task073_commonsenseqa_answer_generation"
+        {
+            "task": "task073_commonsenseqa_answer_generation",
+            "prompt": [
+                "most_suitable_answer",
+				"question_answering",
+				"question_to_answer_index"
+            ]
+        }
     ],
     "dream": [
-        "task247_dream_answer_generation"
+        {
+            "task": "task247_dream_answer_generation",
+            "prompt": [
+                "baseline",
+				"read_the_following_conversation_and_answer_the_question"
+            ]
+        }
     ],
     "quail": [
-		"task887_quail_answer_generation"
+        {
+            "task": "task887_quail_answer_generation",
+            "prompt": [
+                "context_description_question_text",
+				"context_question_description_text",
+				"description_context_question_text"
+            ]
+        }
     ],
     "quartz": [
     ],
     "social_i_qa": [
-        "task384_socialiqa_question_classification",
-		"task580_socialiqa_answer_generation"
+        {
+            "task": "task384_socialiqa_question_classification",
+            "prompt": [
+				"Check if a random answer is valid or not"
+            ]
+        },
+        {
+            "task": "task580_socialiqa_answer_generation",
+            "prompt": [
+				"Show choices and generate answer",
+				"Show choices and generate index"
+            ]
+        }
     ],
     "wiqa": [
     ],
     "cosmos_qa": [
-        "task024_cosmosqa_answer_generation"
+        {
+            "task": "task024_cosmosqa_answer_generation",
+            "prompt": [
+				"context_description_question_text",
+				"context_question_description_text",
+				"description_context_question_text"
+            ]
+        }
     ],
     "qasc": [
     ],
     "quarel": [
-        "task1378_quarel_correct_answer_generation"
+        {
+            "task": "task1378_quarel_correct_answer_generation",
+            "prompt": [
+				"choose_between",
+				"do_not_use",
+				"heres_a_story",
+				"logic_test",
+				"testing_students"
+            ]
+        }
     ],
     "sciq": [
-        "task591_sciq_answer_generation"
+        {
+            "task": "task591_sciq_answer_generation",
+            "prompt": [
+				"Direct Question (Closed Book)"
+            ]
+        }
     ],
     "wiki_hop/original": [
     ],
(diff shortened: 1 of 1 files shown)


doc/prompt_task_map.json
```diff
@@ -0,0 +1,130 @@
...
+        "task247_dream_answer_generation"
+    ],
+    "quail": [
+		"task887_quail_answer_generation"
+    ],
+    "quartz": [
+    ],
+    "social_i_qa": [
+        "task384_socialiqa_question_classification",
+		"task580_socialiqa_answer_generation"
```
danyaljj: Hmm ... isn't `social_i_qa` a question answering task?

danyaljj: if so, I am confused about why it is mapped to an "answer_generation" task.

yeganehkordi: Yes, I mapped our tasks to the datasets. They have created the following tasks types from social_i_qa:
- answer verification (task384_socialiqa_question_classification)
- multiple-choice question answering (task580_socialiqa_answer_generation)
- contextual question answering without options (missing task)
- question generation from the given context and answer (missing task)
(They have one prompt for answer with option index and one for answer with the option string. I just add one of them as a missing task.)

danyaljj: Could you share the pointers to these? 
> answer verification (task384_socialiqa_question_classification)
> multiple-choice question answering (task580_socialiqa_answer_generation)

yeganehkordi: Yeah, answer verification (task384_socialiqa_question_classification) -> [data in sheet](https://docs.google.com/spreadsheets/d/1wo637vMdwa-gEKT9T1eotOXSZiIW1rWyRJzyZcA73sc/edit#gid=1761964683&range=A994), [task](https://github.com/allenai/natural-instructions/blob/master/tasks/task384_socialiqa_question_classification.json)
multiple-choice question answering (task580_socialiqa_answer_generation) -> [data in sheet](https://docs.google.com/spreadsheets/d/1wo637vMdwa-gEKT9T1eotOXSZiIW1rWyRJzyZcA73sc/edit#gid=1761964683&range=A1189), [task](https://github.com/allenai/natural-instructions/blob/master/tasks/task580_socialiqa_answer_generation.json)
You can see the prompts in [Hosted version of PromptSource](https://bigscience.huggingface.co/promptsource).

danyaljj: Not our task; I am asking for the pointers to the prompt-source equivalent of these tasks.

yeganehkordi: Sorry, you can see the prompts in the [hosted version of PromptSource](https://bigscience.huggingface.co/promptsource).
or [here](https://github.com/bigscience-workshop/promptsource/tree/main/promptsource/templates/social_i_qa)

yeganehkordi: "Check if a random answer is valid or not" prompt is answer verification and "Show choices and generate answer" prompt is multiple-choice question answering.

danyaljj: Got it

danyaljj: Follow-up: so, does your script produce several task json files for [this](https://github.com/bigscience-workshop/promptsource/tree/main/promptsource/templates/social_i_qa) prompt file?

yeganehkordi: Yes, it produces 6 task json files (One task for each prompt in the template).

danyaljj: Okay. How can we (i.e., @yizhongw ) know which of these tasks (json file) correspond to `socialiqa_question_classification` or `socialiqa_question_classification`?

Palipoor:
Hi! I am sorry for my absence @danyaljj @yeganehkordi.
This list looks good, do we have a list of task types that we have and they don't? (e.g. task246_dream_question_generation)

doc/prompt_task_map.json
yeganehkordi: I've added the prompt name and id as a key in the json files. 
Also, I can change this file to a prompt to task map.

yeganehkordi:
> Hi! I am sorry for my absence @danyaljj @yeganehkordi. This list looks good, do we have a list of task types that we have and they don't? (e.g. task246_dream_question_generation)

Yes, The list of the missing tasks is in the spreadsheet (based on the dataset and prompt name).
A list of the tasks that we have is in this Json file.

doc/prompt_task_map.json
danyaljj: > I've added the prompt name and id as a key in the json files.

Maybe I'm missing something here. Does the name of these json files tell us whether they correspond to `socialiqa_question_classification` or  `socialiqa_question_classification`?


My understanding is that, here we have a 1-to-2 mapping (as opposed to 1-to-1 mapping). It's possible that I am missing something, in which case, help me see it! :)

yeganehkordi: No, you're right! I think I can use better architecture here. 
Script gets the name of the dataset and generates a task for each prompt, and the name of the json files is based on the dataset and prompt names (e.g., task_socialiqa_Generate_answer).
The mapping is based on the datasets and our tasks and it hasn't mapped the tasks to the prompts. So, I need to add the prompt-task correspondence.

yizhongw:
@yeganehkordi Question: are you including only the T0 training datasets here? It seems some common datasets (e.g., COPA, BoolQ, WSC, etc.) are missing. Those are used for training T0p, T0pp as well as for evaluation. Could we add them?

doc/prompt_task_map.json
danyaljj: Yeah exactly.

yeganehkordi:
> @yeganehkordi Question: are you including only the T0 training datasets here? It seems some common datasets (e.g., COPA, BoolQ, WSC, etc.) are missing. Those are used for training T0p, T0pp as well as for evaluation. Could we add them?

Yes, They are only T0 training datasets. I'll add T0p and T0pp training datasets.

yizhongw:
> Yes, They are only T0 training datasets. I'll add T0p and T0pp training datasets.

Thanks! And also the test tasks mentioned [here](https://huggingface.co/bigscience/T0pp#evaluation-data). We have their mapping before in our working doc, but it's not in a prompt-to-task format.

yeganehkordi:
> > Yes, They are only T0 training datasets. I'll add T0p and T0pp training datasets.
> 
> Thanks! And also the test tasks mentioned [here](https://huggingface.co/bigscience/T0pp#evaluation-data). We have their mapping before in our working doc, but it's not in a prompt-to-task format.

Will do.

doc/prompt_task_map.json
yeganehkordi: I've updated the file to map the tasks to the prompt. I'll add T0p and T0pp datasets tmrw.

doc/prompt_task_map.json
```diff
@@ -0,0 +1,746 @@
...
+				"plain_background_situation",
+				"plain_bottom_hint",
+				"prompt_beginning",
+				"prompt_bottom_hint_beginning",
+				"prompt_mix",
+				"read_background_situation"
+            ]
+        }
+    ],
+    "hotpot_qa/distractor": [
```
yizhongw, line 166: T0 casts hotpot_qa as a closed_book qa task. Can we add a `hotpot_qa/closed_book` placeholder here, and add the task together with other missing tasks?

doc/prompt_task_map.json
```diff
@@ -0,0 +1,746 @@
...
+    ],
+    "winogrande/winogrande_s": [
+    ],
+    "winogrande/winogrande_m": [
+    ],
+    "winogrande/winogrande_l": [
+    ],
+    "winogrande/winogrande_debiased": [
+    ],
+    "story_cloze/2016": [
```
yizhongw, line 725: @yeganehkordi Do you know the differences between the 2016 and 2018 configs? Does T0 use the 2016 config?

doc/prompt_task_map.json
```diff
@@ -0,0 +1,746 @@
...
+    ],
+    "winogrande/winogrande_xs": [
+    ],
+    "winogrande/winogrande_s": [
+    ],
+    "winogrande/winogrande_m": [
+    ],
+    "winogrande/winogrande_l": [
+    ],
+    "winogrande/winogrande_debiased": [
```
yizhongw: For winogrande, could you confirm that the xl/xs/s/m/l/debiased settings only differ in the training sizes? I think we can remove the other and only keep this debiased setting.

doc/prompt_task_map.json
```diff
@@ -0,0 +1,746 @@
...
+				"based on the previous passage",
+				"could you tell me…",
+				"exam",
+				"exercise",
+				"valid_binary",
+				"yes_no_question"
+            ]
+        }
+    ],
+    "super_glue/copa": [
```
yizhongw, line 557: Does `task828_copa_commonsense_cause_effect` or `task827_copa_commonsense_reasoning` correspond to this? We have been using these two tasks for our current evaluation.

doc/prompt_task_map.json
```diff
@@ -0,0 +1,746 @@
...
+            "prompt": [
+				"Direct Question (Closed Book)"
+            ]
+        }
+    ],
+    "wiki_hop/original": [
+    ],
+    "wiki_hop/masked": [
+    ],
+    "adversarial_qa/adversarialQA": [
```
yizhongw, line 86: I think it's fine to keep all the four settings for `adversarial_qa` here. But when we add the missing tasks later, please just use this adversarialQA setting. Or, probably we can just drop the other three now to avoid confusion?

doc/prompt_task_map.json
```diff
@@ -0,0 +1,746 @@
...
+        {
+            "task": "task024_cosmosqa_answer_generation",
+            "prompt": [
+				"context_description_question_text",
+				"context_question_description_text",
+				"description_context_question_text"
+            ]
+        }
+    ],
+    "qasc": [
```
yizhongw, line 60: We have qasc tasks right? `task040_qasc_question_generation`, `task041_qasc_answer_generation`.

doc/prompt_task_map.json
```diff
@@ -0,0 +1,746 @@
...
+            "task": "task580_socialiqa_answer_generation",
+            "prompt": [
+				"Show choices and generate answer",
+				"Show choices and generate index"
+            ]
+        }
+    ],
+    "wiqa": [
+    ],
+    "cosmos_qa": [
```
yizhongw, line 50: I noticed that you included the question generation tasks for some QA datasets. This is good because I remember the original T0 also uses such prompts. But for some QA datasets, you didn't include the question generation task (e.g., `task023_cosmosqa_question_generation` can be included here for `cosmos_qa`). Can we add all of them if we already have them in our current data?

yizhongw (changes requested):
I think the organization of datasets, NI tasks and PS prompts are much more clear. I left some comments for minor fix. @Palipoor Could you work on the missing datasets?

doc/prompt_task_map.json
yeganehkordi: I've only added the tasks that had an equivalent prompt. In this case, we have a question generation task, but they don't have a question generation prompt for `cosmos_qa`.

doc/prompt_task_map.json
yeganehkordi: Yes, but we don't have any shared task with them.

doc/prompt_task_map.json
yeganehkordi: I don't know the setting of our task and I'm not sure if it is `adversarial_qa` or not. 
If we don't need all the subsets, we can change the instances of this task and just keep `adversarial_qa`.

doc/prompt_task_map.json
yeganehkordi: There is a slight difference between our task and their tasks. Our task is to choose the completion which is the cause or effect of the first sentence, but they have specified that the completion should be which one of them in the instances.
So, I think our task is more general and probably more difficult.

doc/prompt_task_map.json
yeganehkordi: Yes, Their only difference is their sizes. 
Sure, will do.

doc/prompt_task_map.json
yeganehkordi: The data fields are the same among the 2016 and 2018 splits. Instances might be different.
Yes, they have used the 2016 [config for T0](https://github.com/bigscience-workshop/t-zero/blob/2604dd383c72e864b73efb3c5bce19892c7a679c/evaluation/template_list.py#L67).

doc/prompt_task_map.json
yeganehkordi: Can you elaborate more?
They don't have any closed_book prompt. are you suggesting adding this task without prompt?

yizhongw: It seems they used the [kilt version of hotpotqa](https://huggingface.co/datasets/kilt_tasks/viewer/hotpotqa/train), according the the list [here](https://github.com/bigscience-workshop/t-zero/blob/2604dd383c72e864b73efb3c5bce19892c7a679c/t0/datasets.csv).

doc/prompt_task_map.json
yizhongw: Got it. Are we going to add a task similar to theirs?

doc/prompt_task_map.json
yeganehkordi: Thanks! I'll add it.

doc/prompt_task_map.json
yeganehkordi: Yeah, I will.

danyaljj:
Merging this PR since we iterated over it several times now. If anything is missing let's address in another PR.

PR #746, opened 2022-04-01
Add missing tasks from the prompt source

yeganehkordi:
I've added the last 5 tasks (I had miscalculated them before). I think we covered all the prompt source evaluation datasets with this pr.


Palipoor:
In `task1296_wiki_hop_question_answering` isn't it better to have the context before the options? In this order perhaps:
Context - Subject - Relation - Options.
Other things look good to me! Thanks for correcting the files I added.

yizhongw:
LGTM!

yizhongw (approved):

PR #750, opened 2022-04-02
Drop extra words from the definitions

yeganehkordi:


danyaljj:
Thanks!

danyaljj:
FYI @yizhongw

Issue #752, opened 2022-04-03
Duplicate tasks: 1508 and 1509


todpole3:
Task files in repo are exactly the same. Perhaps the correct file for one of the task is missing.

danyaljj:
@yeganehkordi can you take a look into this?

yeganehkordi:
Yeah, They are exactly the same. I will check the files in the previous releases.

danyaljj:
Addressed in #753

PR #483, opened 2021-10-25
task1508-1510 (wordnet antonym generation, evalution antonym generation and evalution relationship extraction), 1582-1585

atharva-naik:
task1508 (synthetic/wordnet): Given an adjective, generate it's antonym. An antonym of a word is a word opposite in meaning to it.
task1509 (EVALution dataset): Given an adjective, generate it's antonym. An antonym of a word is a word opposite in meaning to it.
task1510 (EVALution dataset): Given a phrase describing the relationship between two words, extract the words and the lexical relationship between them.

PR #527, opened 2021-11-01
renamed task 1511-1514 to 1582-1585

atharva-naik:
Renaming of tasks 1511 (bless_hypernym_generation) to 1514 (root09_hypernym_generation) to 1582-1585. All tests running successfully.

PR #753, opened 2022-04-04
Update task1509_evalution_antonyms

yeganehkordi:
Addressing the #752 
These tasks are similar but their sources are different. I restored the previous version of task 1509 in #483.


danyaljj:
@yeganehkordi do we know in what PR the two tasks became identical?

yeganehkordi:
> @yeganehkordi do we know in what PR the two tasks became identical?

Yes, After closing the first PR(#483), they became identical in #527. It was the contributor's mistake and since they were reviewed in the first PR, we didn't notice that.

danyaljj:
I see. Thanks for clarifying!

todpole3:
Thanks very much for addressing this @yeganehkordi and @danyaljj!

Issue #755, opened 2022-04-05
Ungrammatical Instructions for Some Tasks


todpole3:
Task575 - "The input is taken a dialogue from the conversation between an agent and the customer. Output Agent if the dialogue is spoken by the agent in the conversation, otherwise output customer if it is spoken by the Customer."

Task1599 - "In this task, you have given an utterance, if it's user_utterance label it 1)User and if it's agent_utterance label it 2)Agent. The user always gives commands to the agent based on scheduling events, retrieving event information, and checking the schedule. While Agents' reply is answer of users' question or follow-up question on users' command."

I figure it might be helpful to bring this to attention. Maybe the dataset should tolerate some noise though...

yeganehkordi:
Thanks for your attention, I'll fix them.

PR #756, opened 2022-04-05
Update task 575 and task 1599

yeganehkordi:
Addressing #755


todpole3:
Thanks for addressing this!

PR #758, opened 2022-04-12
Create a single file for crowd annotation using the tasks in test categories.

yizhongw:


eval/amt/1.amti_create_input_data.py
```diff
@@ -84,18 +87,112 @@ def process_single_file(start, end, max_count):
             fout.write(json.dumps(map) + "\n")
 
 
+def process_default_test_tasks(max_count):
```
danyaljj: I think this is maxing over instances (not tasks); right?  if so, let's change the name so that it reflects it (e.g., `max_instance_count`).

danyaljj (approved):
One small suggestion. Otherwise, it looks good!

eval/amt/1.amti_create_input_data.py
yizhongw: Updated.

PR #764, opened 2022-05-09
Add standard evaluation setup

yizhongw:
- Add the train/test splits in `splits` folder. 
- Add a new field `Instances Subset for Testing` in json files of the 154 evaluation tasks.
- Add the README for evaluation setup.


danyaljj:
Thanks! 

I wonder if instead of creating a new field for the 100 subset of instances, we can just say the first 100 instances? If there is a concern regarding reproducibility, we can re-order the instances so that your random evaluation instances appear as the first 100 ones.

yizhongw:
> Thanks!
> 
> I wonder if instead of creating a new field for the 100 subset of instances, we can just say the first 100 instances? If there is a concern regarding reproducibility, we can re-order the instances so that your random evaluation instances appear as the first 100 ones.

Sounds good to me. Let me re-order the instances.

splits/README.md
```diff
@@ -0,0 +1,23 @@
+# Evaluation Setup of Natural Instruction V2
+
```
danyaljj, line 2: Suggested addition: 

"Here we describe the evaluation setup used in [this paper](https://arxiv.org/abs/2204.07705) which can be used for reproducing our experiments or extending them."

splits/README.md
```diff
@@ -0,0 +1,23 @@
...
+To facilitate research on cross-task generalization, we split out tasks into two subsets: one for training and one for testing. Users can use the training tasks for any modeling purpose, while the testing tasks should be used to evaluate the model's ability to solve new tasks by following instruction.
```
danyaljj: Suggested rephrasing: 
"The target setup of the study is a cross-task generalization, i.e., training on a subset of tasks and evaluating on the remaining unseen ones. To do so, we split ... "

splits/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## English Track
```
danyaljj, line 7: These appeared suddenly. I think you need a sentence to connect the splits to "tracks".

src/split_tasks.py
```diff
@@ -0,0 +1,161 @@
+import argparse
+import glob
+import random
+import json 
+import os
+from secrets import choice
+
```
danyaljj, line 14: Add a few sentences describing what this script does.

danyaljj:
I think we should add a few lies to the testing script (https://github.com/allenai/natural-instructions/blob/master/src/test_all.py) in order to make sure that: 
 - Each task appears in the split  
 - East task in the split actually exists 
 - There is no overlap between test and train task names in the splits files.

yizhongw:
> I think we should add a few lies to the testing script (https://github.com/allenai/natural-instructions/blob/master/src/test_all.py) in order to make sure that:
> 
> * Each task appears in the split
> * East task in the split actually exists
> * There is no overlap between test and train task names in the splits files.

sure, will do!

splits/README.md
```diff
@@ -0,0 +1,29 @@
+# Evaluation Setup of Natural Instruction V2
```
danyaljj, line 1: Somewhere you may want to add a reference to your models (https://github.com/yizhongw/Tk-Instruct)

src/test_all.py
```diff
@@ -278,6 +278,21 @@ def natural_keys(text):
...
+    excluded_tasks = [l.strip() for l in open(f"splits/{split_name}/excluded_tasks.txt")]
+
+    # make east task in the split actually exists
+    for task in train_tasks + test_tasks + excluded_tasks:
+        assert task in task_names, f" Task {task} doesn't exist, but it's included in the {split_name} split."
+    # make sure each task appears in the split 
+    for task in task_names:
+        assert task in train_tasks + test_tasks + excluded_tasks, f" Task {task} is missing in the {split_name} split."
+    # make sure there is no overlap between test and train task names in the splits files.
+    assert len(set(train_tasks) & set(test_tasks)) == 0, f" {split_name} split has overlap tasks in the train & test sets."
```
danyaljj, line 294: Do the same check with `excluded`?

splits/README.md
danyaljj, line 1: Reference this file from the main readme file?

danyaljj: Add a reference to TK-instruct repo to the main readme too!

PR #766, opened 2022-05-30
Fix arg name at rouge

lkm2835:
Fixed argument name `use_agregator` to `use_aggregator`

Reference: https://github.com/huggingface/datasets/blob/master/metrics/rouge/rouge.py#L65


danyaljj:
Thanks! 

FYI: @yizhongw

